PR #50, opened 2017-06-27
Improve HiDPI visuals

vjr:
This is an initial (and very ugly I'm sure) attempt to improve the greeter visual behaviour with hidpi enabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,6 +84,7 @@ PACKAGES
     wingpanel-2.0
 OPTIONS
     --vapidir=${CMAKE_CURRENT_SOURCE_DIR}/vapi
+    --thread
 )
 
 add_subdirectory (po)
--- a/src/PantheonGreeter.vala
+++ b/src/PantheonGreeter.vala
@@ -61,29 +61,24 @@ public class PantheonGreeter : Gtk.Window {
     private string state_file;
 
     public static PantheonGreeter instance { get; private set; }
-    private static SettingsDaemon settings_daemon;
 
     //from this width on we use the shrinked down version
     const int NORMAL_WIDTH = 1200;
     //from this width on the clock wont fit anymore
     const int NO_CLOCK_WIDTH = 920;
 
-    public static bool TEST_MODE { get; private set; }
+    public static bool TEST_MODE;
 
     public PantheonGreeter () {
         //singleton
         assert (instance == null);
         instance = this;
 
-        TEST_MODE = Environment.get_variable ("LIGHTDM_TO_SERVER_FD") == null;
-
         if (TEST_MODE) {
             message ("Using dummy LightDM because LIGHTDM_TO_SERVER_FD was not found.");
             login_gateway = new DummyGateway ();
         } else {
             login_gateway = new LightDMGateway ();
-            settings_daemon = new SettingsDaemon ();
-            settings_daemon.start ();
         }
 
         var state_dir = Path.build_filename (Environment.get_user_cache_dir (), "unity-greeter");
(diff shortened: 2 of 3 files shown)


vjr:
So to start of , if you see the code, I have used a couple of `Gtk.main_iteration_do (true);` calls because for whatever reason the code seems to "get stuck" - signals don't get invoked not does the `watch_name` call seem to execute. What am I doing wrong?

I also tried setting the `--thread` option to valac in the cmake file as you can see but that didnt seem to help.

Also the second callback via the session proxy handle is never invoked, so the second wait method just continues after a 1 second delay and luckily by that time the scale factor is available before Clutter Init.

davidmhewitt:
I'm not sure if this is possible, but it might be worth a try:

We know that the scale factor information is contained within the GSD xsettings plugin. So if there was a way to reliably detect whether that plugin was ready or not yet, it would be much better than waiting an arbitrary 1 second and hoping it was there. This may vary depending on people's hardware.

So maybe try watching the `org.gnome.settings-daemon.plugins.xsettings` bus for owners or property changes? If that's possible, then we'll reliably know if the bus is ready or not.

vjr:
@davidmhewitt I'm already attempting to listen for the xsettings plugin to load but the signal never seems to arrive, my latest commit, i increased the wait time to 3 seconds but because the signal does not seem to arrive the login screen takes about 3 seconds to show up. even though the signal does not arrive in the code the greeter clutter seems to pick up the scale fine meaning GSD and the plugin are actually running by the time the 3 seconds is up.

Can you see what's wrong/missing in the code or maybe config (cmake) that is causing the signal to not arrive? I actually suspect GSD itself is not sending the signal but I'm reading its code now.

So if we can't figure out how to actually get the signal, then the small hardcoded delay (like 1-3 seconds) happens which I suspect users will complain about :-)

davidmhewitt:
@vjr You're listening on the `org.gnome.SettingsDaemon` bus. I meant it might be worth a try to listen for changes on the `org.gnome.settings-daemon.plugins.xsettings` bus. As I think some changes should happen on that bus when the plugin is loaded.

vjr:
Just a note that this PR is to attempt to address issue #7

vjr:
Looking at the correct GSD version 3.18.2 code for loki (from ubuntu xenial package sources) there may be potential problems in the code such that the plugin_activated signal is never sent over dbus even though the code is there to do it. See the FIXME comment in [the code around here](https://git.gnome.org/browse/gnome-settings-daemon/tree/gnome-settings-daemon/gnome-settings-manager.c?h=gnome-3-18#n138). For example, notice the `const char * signal` parameter is not used and the signal name is hardcoded to `PluginActivated` which means even the `on_plugin_deactivated` callback will emit the activated signal !

Right now I'm going to try building GSD locally with additional logging statements to see if at all the activated signal is being emitted, which I think may not be the case. If this is so, then this greeter hidpi issue PR will need to be reworked a bit - maybe just wait to connect to the GSD dbus name and then either wait for an arbitrary time like 1 second.

Other options? Some other way to wait for GSD xsettings plugin to be ready?

vjr:
Yep, GSD never sends the signals for the FIXME reason mentioned in the previous comment, the dbus connection is NULL (is ready a little later on) so the `emit_signal` function returns early. I got the signals to work by changing the use of `g_bus_get` to `g_bus_get_sync` so GSD is ready with dbus before trying to load plugins and sending the signals. But I don't think it's wise to mess with GSD code to avoid any risks.

If the greeter hidpi visual glitch is to be improved (if not totally fixed) I'd suggest just waiting for the GSD dbus name to appear (which is already in the PR code) and then optionally waiting for about 1 second. I can try just doing the first wait (for the dbus name) and not the arbitrary subsequent 1 second wait to see if it works. It will anyway likely be flaky (still exhibit the current glitch) possibly on older/slower computers.

Note the code still has the ugly `Gtk.main_iteration_do ()` calls which I need some comments on how to avoid it and clean up the code.

vjr:
Second attempt with a much smaller diff here: https://github.com/elementary/greeter/pull/51

vjr:
Closing this as it's superseded by another branch via https://github.com/elementary/greeter/pull/51

PR #51, opened 2017-06-28
Improve HiDPI visuals - take 2 - simplified

vjr:
Now that it's confirmed GSD fails to signal when its plugins (like xsettings) are activated, as noted in PR #50 in [this comment](https://github.com/elementary/greeter/pull/50#issuecomment-311472052), here's a much simplified patch to try and make greeter look good on hidpi displays!

I'm still concerned about the need to call `Gtk.main_iteration_do (true)` and the 1 second sleep is needed to allow (with fingers crossed) the xsettings GSD plugin to take effect (and thus the scaling-factor to be available) before clutter init.

Hopefully this is good enough to at least get into Loki daily/testing PPA - I am really enjoying the improved hidpi look and would love to see this in Loki public stable update for regular users!


vjr:
Just a note that this is to address issue #7

matzipan:
Does this supersede #51? If so, can you close it? Andrei

vjr:
You mean does this supersede #50 ? Which it does, so I closed that one. Thanks!

vjr:
Added the blocked status because this branch now requires GSD to be patched to actually send the `plugin_activated` signals. There's a patched package in my PPA here: https://launchpad.net/~vishalrao/+archive/ubuntu/elementaryos and this branch works well with it.

davidmhewitt:
I imagine this is no longer relevant now that the Juno greeter runs under a basic WM. It also has conflicting files so I'll close this and it can be re-opened at a later date if it's still needed for something.

Issue #52, opened 2017-06-28
Time isn't displayed in 24h format, despite setting in switchboard


jtrees:
In the System Settings, I set the time format to 24h but greeter ignores this.

jtrees:
Related: https://github.com/elementary/greeter/issues/17

vjr:
The correct PR link is https://github.com/elementary/greeter/pull/53

jtrees:
Cool thanks. I look forward to this being merged.

richo67:
When will this be merged? When we can see that in use? Thanks

jlnr:
Besides using the wrong format, the time string is also clipped on German systems now (running at 4K):

![greeter](https://user-images.githubusercontent.com/70772/40164810-d5e4cd60-59ba-11e8-9a00-41a9c81d5875.jpg)

I think this is new, and it used to say "vorm." instead of "vormittags".

danrabbit:
Closing as a duplicate of #17

PR #53, opened 2017-06-29
Show greeter time format using clock format setting of selected user.

vjr:
This is to address issue https://github.com/elementary/greeter/issues/52


matzipan:
Can you briefly explain the issue, steps taken and why?

src/LoginOptions/UserLogin.vala
```diff
@@ -22,9 +22,34 @@ public class UserLogin : LoginOption {
...
+
+        string gsettings_result;
+        string? xdg_config_home_real = Environment.get_variable (UserLogin.ENV_VAR_NAME);
+        string xdg_config_home_spoof = Path.build_path (Path.DIR_SEPARATOR_S,
+                                        user.home_directory, UserLogin.ENV_VAR_SUFFIX);
+
+        Environment.set_variable (UserLogin.ENV_VAR_NAME, xdg_config_home_spoof, true);
+
+        try {
+            Process.spawn_command_line_sync ("gsettings get org.gnome.desktop.interface clock-format",
```
matzipan: I see here you're using a GSettings command line spawn to read the setting. Perhaps a better approach would be to do this programatically, instead?

src/Widgets/TimeLabel.vala
```diff
@@ -46,19 +47,26 @@ public class TimeLabel : Gtk.Grid {
         show_all ();
     }
 
+    public void set_format (string? clock_format) {
```
matzipan: Shouldn't this use a Vala setter pattern instead?

src/Widgets/TimeLabel.vala
```diff
@@ -46,19 +47,26 @@ public class TimeLabel : Gtk.Grid {
...
+
     bool update_time () {
         var date = new GLib.DateTime.now_local ();
 
         /// Date display, see http://valadoc.org/#!api=glib-2.0/GLib.DateTime.format for more details
         var day_format = _("%A, %B %e");
         /// Time display, see http://valadoc.org/#!api=glib-2.0/GLib.DateTime.format for more details
         var time_format = _("%l:%M");
+        /// Time display, see http://valadoc.org/#!api=glib-2.0/GLib.DateTime.format for more details
+        var time_format_24h = _("%H:%M");
```
matzipan: Can you make use of the [Granite utilities](https://github.com/elementary/granite/blob/1def8c65945f9f83acf94ae6e58897d083c7227b/lib/Widgets/Utils.vala)? Especially since they are already translated and everything.

src/LoginOptions/UserLogin.vala
vjr: When I tried the shorter way of using `Granite.Services.Settings` or plain `GLib.Settings` I wasn't able to get them to pick up individual users' settings. No way to trick them by setting the `XDG_CONFIG_HOME` env variable either. Another option was to copy the GNOME gvdb low level C code (and write vapi files) to read the binary settings file for every user but that is a bigger task and maybe not really ideal either.

src/Widgets/TimeLabel.vala
vjr: This method also calls the `update_time ()` method which is why I thought not to make it a setter property.

src/Widgets/TimeLabel.vala
vjr: Ah, I see now there is `get_default_time_format ()` in Granite library. OK I will try it just for the new var I added, but note the existing ones don't seem to use it. Will try it and update the PR if it works. Thanks!

vjr:
@matzipan here's a note about what is the issue, steps taken and why:

The issue is that I've seen a few users complain/ask about the greeter timeformat being always showing "AM / PM " 12h format. They have expressed a "wishlist" level request like issue https://github.com/elementary/greeter/issues/52 where they want the time to be in the format that user has set in their account (via switchboard plugin datetime). Now I've seen some responses from other folks like "what if there are multiple users with different settings" or "greeter cannot access the per-user setting" and even if it could, which one to show, given more than one user with different settings.

So the approach I took was to change the timelabel format to show whatever the current user has set when the user selection changes, such as by scrolling the mouse. By doing this, if there is only one user the greeter will show that user's preferred time format and if there are multiple users, the greeter already pre-selects the last logged in user, so if a computer is mostly used by only one person they will also see their preferred setting on the greeter login screen.

I had to take the approach of using `Process.spawn...` call because the more direct APIs were not able to access the per-user setting as I wrote here: https://github.com/elementary/greeter/pull/53#discussion_r124932460

src/Widgets/TimeLabel.vala
matzipan: Well a setter can call any functions inside of it.

matzipan:
> So the approach I took was to change the timelabel format to show whatever the current user has set when the user selection changes, such as by scrolling the mouse. By doing this, if there is only one user the greeter will show that user's preferred time format and if there are multiple users, the greeter already pre-selects the last logged in user, so if a computer is mostly used by only one person they will also see their preferred setting on the greeter login screen.

I think this part will require @danrabbit to chip in.

src/LoginOptions/UserLogin.vala
```diff
@@ -22,9 +22,34 @@ public class UserLogin : LoginOption {
...
+                                                out gsettings_result);
+        } catch (SpawnError e) {
+            debug ("Could not spawn gsettings: %s", e.message);
+        }
+
+        this.clock_format = gsettings_result;
```
matzipan: As per coding guideline, we dont' write `this.`. Furthermore, if a SpawnError exception is triggered, what will the value of clock_format be?

src/Widgets/TimeLabel.vala
matzipan: As discussed, I think this should be a setter.

src/Widgets/TimeLabel.vala
```diff
@@ -46,19 +47,26 @@ public class TimeLabel : Gtk.Grid {
...
+        update_time ();
+    }
+
     bool update_time () {
         var date = new GLib.DateTime.now_local ();
 
         /// Date display, see http://valadoc.org/#!api=glib-2.0/GLib.DateTime.format for more details
         var day_format = _("%A, %B %e");
         /// Time display, see http://valadoc.org/#!api=glib-2.0/GLib.DateTime.format for more details
         var time_format = _("%l:%M");
```
matzipan: Could `DateTime` not be used here as well?

src/Widgets/TimeLabel.vala
vjr: I saw in existing code, like Wallpaper.set_wallpaper () is also a method and not a setter property because it does extra work other than just setting a property :-) I believe this is common convention across programming languages?

src/Widgets/TimeLabel.vala
```diff
@@ -46,19 +47,26 @@ public class TimeLabel : Gtk.Grid {
...
+    public void set_format (string? clock_format) {
+        format24h = (clock_format != null && clock_format.contains ("24h"));
+        update_time ();
+    }
+
     bool update_time () {
         var date = new GLib.DateTime.now_local ();
 
         /// Date display, see http://valadoc.org/#!api=glib-2.0/GLib.DateTime.format for more details
         var day_format = _("%A, %B %e");
```
matzipan: Perhaps this should also go into `Granite` and `get_default_date_format` would have an extra parameter called `is_abbreviated`.

vjr: I agree, but I believe that is for another PR for another day :-)

src/Widgets/TimeLabel.vala
vjr: Currently Granite datetime utility does not have the long date format... and the greeter has TIME portion separated from the AM/PM portion. I have another PR to combine the time into a single string which will address this partially.

src/Widgets/TimeLabel.vala
matzipan: I shall open an issue for this.

src/LoginOptions/UserLogin.vala
vjr: fixed.

matzipan:
[Preview useful for UX review](https://goo.gl/photos/CwpeyXFPeXZuPRvJ6), a preview of how the switching looks like. It's a bit annoying that it's not centered on the same like, but I guess it's not really worth fixing.

vjr:
> Final change request from my part:

@matzipan - I've made the changes.

matzipan:
I know see in #54 that the horizontal position of the text is set using this formula:
```
time_actor.x = width - time_actor.width - time_actor.width / 2 - 200 / scale_factor;
```

Do you think you can change this formula so that we can center the time label on a given position, so that it doesn't jump around like in the video I posted above? I tried to play around with it but I couldn't get it to work.

vjr:
@matzipan 

> Do you think you can change this formula so that we can center the time label on a given position, so that it doesn't jump around like in the video I posted above? I tried to play around with it but I couldn't get it to work.

It would have been straightforward if the time actor was left aligned but since it's on the right, I don't at the moment see a way to fix a safe position because of different widths for different users who have either 12h or 24h format set.

matzipan:
Where would the time_actor need to be left aligned? I'm not familar with Clutter.

vjr:
I meant only "if" the timelabel were placed on the left side, like the user login actor, it would have been easier :-) You can see the user login actor horizontal distance is hardcoded (pretty much) unlike the time actor which uses its width in the calculation.

src/Widgets/TimeLabel.vala
```diff
@@ -46,19 +47,28 @@ public class TimeLabel : Gtk.Grid {
...
+        /// Time display, see http://valadoc.org/#!api=glib-2.0/GLib.DateTime.format for more details
+        var time_format_24h = Granite.DateTime.get_default_time_format (false);
         /// AM/PM display, see http://valadoc.org/#!api=glib-2.0/GLib.DateTime.format for more details. If you translate in a language that has no equivalent for AM/PM, keep the original english string.
         var meridiem_format = _(" %p");
 
         date_label.label = date.format (day_format);
-        time_label.label = date.format (time_format);
-        pm_label.label = date.format (meridiem_format);
+        time_label.label = date.format (format_24h ? time_format_24h : time_format);
+        pm_label.label = date.format (format_24h ? "" : meridiem_format);
```
danrabbit: Can we use a Gtk.Revealer here instead with a slide animation so that it's not so jarring to switch between users with different formats? I guess theoretically that shouldn't happen, but I feel like we should handle it more gracefully

matzipan: This would be a workaround for the issue I raised above with the alignment, so :+1:

vjr: So I'm not familiar with UI coding but my latest commit https://github.com/elementary/greeter/pull/53/commits/372e96ebb039bdb6c9ad8f63d1f5a746e664534a adds a fadein-fadeout animation with the actual format switch when the time is invisible to avoid seeing the flash. I couldn't figure out how to get the revealer working for either slide or crossfade so I followed the existing clutter actor animation code.

Here's a video of it in action: https://1drv.ms/v/s!AHyYMRNCqMZwlGE

What do you folks think? (Of the code and the choice of animation itself).

vjr: @danrabbit  what do you think?

Philosoft:
oh. c'mon guys. It's been 5 freaking years since [this](https://bugs.launchpad.net/pantheon-greeter/+bug/1048873) and still half a year passed since last activity... it's just sad

jlnr:
Besides using the wrong locale, the AM/PM string is also clipped on German systems now:

greeter

(I think this is new, and it used to say "vorm." instead of "vormittags". But I may be wrong)

Is the only remaining blocker here the decision between Gtk.Revealer and the fading in the OneDrive video?

jlnr:
I've just tried this branch. This apparently picks up the date format of the selected user, but not their language. If I select a US user on my German machine, it still displays the German "10:58 NACHMITTAGS", but it should be the English "10:58 PM".

danrabbit:
Gonna close this since it now conflicts with master pretty severly

Issue #17, opened 2017-04-12
better time format support in login screen


elementaryBot:
After installing elementaryOS Loki in Chinese (Taiwan) language, I encountered 2 problems with the time format in the login screen. Please see the attached screen photo for details.

1. Chinese (Taiwan) locale uses "%p%l:%M" for 12hr format, but translators can only translate "%l:%M" and " %p" separately in Launchpad which cannot deal with the format we use. [1] The better way to do it is to let translators translate "%l:%M %p" as a whole not separately.

2. The time string in Chinese is cut in a small screen, the resolution is 1366x768. Please see the attachment.

[1] https://translations.launchpad.net/pantheon-greeter/trunk/+pots/pantheon-greeter/zh_TW/+translate

Launchpad Details: [#LP1624217](https://bugs.launchpad.net/bugs/1624217) Cheng-Chia Tseng - 2016-09-16 05:18:39 +0000

elementaryBot:
I am working in my environment to fix not only this issue but also 1639845. So far it seems to me that solution is in combination of proper positioning of label displaying time, formating of the time CSS (making font a bit smaller in apps.css as well as using time format %X.

I wonder how did you make the screen shot of the login screen?

Launchpad Details: [#LPC](https://bugs.launchpad.net/pantheon-greeter/+bug/1624217/comments/2) richo - 2017-01-04 17:48:52 +0000

elementaryBot:
I am working in my environment to fix not only this issue but also 1639845. So far it seems to me that solution is in combination of proper positioning of label displaying time, formating of the time CSS (making font a bit smaller in apps.css as well as using time format %X.

I wonder how did you make the screen shot of the login screen?

Launchpad Details: [#LPC](https://bugs.launchpad.net/pantheon-greeter/+bug/1624217/comments/3) richo - 2017-01-08 19:16:43 +0000

vjr:
Correct link to the PR is https://github.com/elementary/greeter/pull/54

vadbars:
An error occurs in elementary OS 5.0 Juno. The same as a year ago in Loki. )

kr428:
Still valid in 5.0 it seems. :(

peteruithoven:
Could someone verify that this is fixed with the new (unreleased) Greeter?

Looks like the Greeter uses Granite for the date and time formatting:
[...]

PR #54, opened 2017-06-30
Address i18n issue with greeter timelabel.

vjr:
This is to address the first point in the issue https://github.com/elementary/greeter/issues/17 by combining the time and am/pm labels into a single text label format.

One nitpick I have is that the resulting AM/PM portion becomes larger where I preferred the current styling.

Note that I've removed the am/pm `meridien_format` label hence that string is removed from the code. Need to update the translations files which I have yet to figure out and add to this PR.


src/Widgets/TimeLabel.vala
```diff
@@ -52,13 +47,10 @@ public class TimeLabel : Gtk.Grid {
         /// Date display, see http://valadoc.org/#!api=glib-2.0/GLib.DateTime.format for more details
         var day_format = _("%A, %B %e");
         /// Time display, see http://valadoc.org/#!api=glib-2.0/GLib.DateTime.format for more details
-        var time_format = _("%l:%M");
-        /// AM/PM display, see http://valadoc.org/#!api=glib-2.0/GLib.DateTime.format for more details. If you translate in a language that has no equivalent for AM/PM, keep the original english string.
-        var meridiem_format = _(" %p");
-
+        var time_format = Granite.DateTime.get_default_time_format (true);
```
matzipan: Perhaps add a new line here.

tintou:
Requested elementary/ux review because of the meridian label removal but it should be fine

matzipan:
Ok so, testing on my machine, when I select another username in the list which has 24h instead of AM/PM set, the format doesn't switch.

vjr:
> I don't know how much this is a styling error, this might be due to the AM/PM label getting to big as you mentioned, but the right hand side of the label is now exactly positioned against the right edge of the screen. Not sure where the change needs to take place.

Should be fixed in the latest commit https://github.com/elementary/greeter/pull/54/commits/05954ab7f300765938f10e65d2728bc793eed490

> Similarly - not entirely sure where this is coming from or if this is even related to this branch, but if I run ./pantheon-greeter or pantheon-greeter with the code on this branch, I get 24h format, but if I lock my screen I get the 12h format.

This may be because you have set your region to UK only in your own account but not in system-wide locale (in switchboard language-region settings) and I believe I also see this behaviour. This should be resolved with the other PR #53

vjr:
Would anyone from @elementary/ux like to take a look and merge this approved PR?

matzipan:
Hey @vjr can you action the last comment I made?

vjr:
@matzipan apologies for the long delay - i will look into this and update (ASAP).

cassidyjames:
Closing due to inactivity. Feel free to re-open if work picks back up!

PR #57, opened 2017-07-02
Fix wallpaper login sizing glitch with multi monitors.

vjr:
It was way back when we had fixed the greeter wallpaper sizing glitch during login on hidpi here https://github.com/elementary/greeter/commit/0339a4e3eef38cd9218dc0aa72668827c9dad03d

A similar sizing glitch during login still occurs with multiple monitors, which this patch intends to fix.

Tested on dual monitor (second one being a TV connected via hdmi) with and without hidpi enabled and with and without the second monitor connected, no glitches observed!

Note the second monitor still has a wallpaper sizing glitch after login but I guess that might be a gala/mutter issue because the greeter does not draw on any monitor other than the primary one IINM.

At least users no longer have to watch the wallpaper glitch while looking at the login screen to desktop transition.


src/PantheonGreeter.vala
```diff
@@ -397,8 +401,12 @@ public class PantheonGreeter : Gtk.Window {
         ctx.set_source_rgba (0.0, 0.0, 0.0, 0.0);
 
         int scale_factor = get_screen ().get_root_window ().get_scale_factor ();
-        int width = get_screen ().get_width () * scale_factor;
-        int height = get_screen ().get_height () * scale_factor;
+
+        Gdk.Rectangle geometry;
+        get_screen ().get_monitor_geometry (get_screen ().get_primary_monitor (), out geometry);
```
cassidyjames: It looks like you're defining this twice, any way to combine that?

vjr: It's already defined locally elsewhere in existing code. I can define it at a starting point but then would need to change a few method signatures to include the `geometry` parameter. Should I do it? I prefer not to, because it looks less neat IMHO.

src/PantheonGreeter.vala
```diff
@@ -380,17 +380,21 @@ public class PantheonGreeter : Gtk.Window {
         unowned X.Display display = gdk_display.get_xdisplay ();
 
         var root_window = (screen.get_root_window () as Gdk.X11.Window);
+
+        Gdk.Rectangle geometry;
+        get_screen ().get_monitor_geometry (get_screen ().get_primary_monitor (), out geometry);
```
donadigo: You're calling `get_screen ()` twice here, you should probably optimize that to:
```
var screen = get_screen ();
screen.get_monitor_geometry (screen.get_primary_monitor (), out geometry);
```

src/PantheonGreeter.vala
```diff
@@ -401,8 +405,12 @@ public class PantheonGreeter : Gtk.Window {
         ctx.set_source_rgba (0.0, 0.0, 0.0, 0.0);
 
         int scale_factor = get_screen ().get_root_window ().get_scale_factor ();
-        int width = get_screen ().get_width () * scale_factor;
-        int height = get_screen ().get_height () * scale_factor;
+
+        Gdk.Rectangle geometry;
+        get_screen ().get_monitor_geometry (get_screen ().get_primary_monitor (), out geometry);
```
donadigo: The same thing happens here.

src/PantheonGreeter.vala
donadigo: I think what you should do here, is to just make a new method that will return you the desired geometry / size for the right monitor like this:

```
private void get_primary_monitor_geometry (out Gdk.Rectangle geometry) {

}
```

Or even simpler:
```
private void get_primary_monitor_size (out int width, out int height) {

}
```

matzipan:
Hey @vjr  any updates on this pull request?

vjr:
@matzipan apologies for the long delay - i will look into this and update (ASAP).

danrabbit:
Gonna close this since it conflicts with master. Feel free to reopen if that changes :)

PR #60, opened 2017-07-08
Fix showing wrong user wallpaper during login

vjr:
This one-liner fixes the case where, during login, the greeter briefly displays the wallpaper of the last highlighted user instead of the current user being logged in.

This issue occurs when you scroll to different users with different wallpapers in the greeter list, it only saves the last un-cached wallpaper for display during the login transition. So the result is that during login you first see the expected wallpaper of the user, then for a brief moment you see the other wallpaper displayed, after that it again switches to the expected user wallpaper when gala starts up.

Of course to notice this the wallpapers themselves have to be different images in the first place.


vjr:
@cassidyjames or @Philip-Scott would either of you do the honours of merging this please? Test it if you like, too !

vjr:
Thanks @danrabbit for the reminder over Slack. So would either @Philip-Scott or @cassidyjames please merge this? It's already approved and ready to click the green merge button :-)

PR #82, opened 2017-12-11
Update GSD processes

### LINKED ISSUE #99: Font in shut down menu and dialog box not Open Sans
It looks like the font in the shut down menu and consequent dialog box are not Open Sans. I don't actually know what they are, and I have no way of confirming this because I don't possess the skills to figure out how to do it. @cassidyjames strongly encouraged me to file this bug report myself, so here it is in all its amateur glory. Thanks for your time.

danrabbit:
This updates Greeter to launch the needed GSD processes in Juno

Fixes #99 

As an aside, this does not fix HiDPI. Greeter will now require a mutter compositor in order to fix this and that is outside the scope of this branch

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,15 +19,6 @@ add_definitions (-DGETTEXT_PACKAGE=\"${GETTEXT_PACKAGE}\")
 # Comment this out to enable C compiler warnings
 add_definitions (-w)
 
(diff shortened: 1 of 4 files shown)


cassidyjames:
This **does** fix #99 fwiw

Before:
greeter-dialog-before
Aftrer:
greeter-dialog-after

PR #93, opened 2018-03-25
Get wallpaper manually if LightDM doesn't work

### LINKED ISSUE #88: Login screen wallpaper isn't changed
When I change my user's wallpaper, wallpaper on login screen doesn't changed.

I make some debug. In this line https://github.com/elementary/greeter/blob/master/src/PantheonGreeter.vala#L199 `userlist.current_user.background` is null. So, looks like somewhere something did not written.

UPD: I am on Juno

mdh34:
works around LightDM returning null background strings in newer versions
fixes #88

### DIFF
--- a/src/LoginOptions/UserLogin.vala
+++ b/src/LoginOptions/UserLogin.vala
@@ -36,7 +36,10 @@ public class UserLogin : LoginOption {
                 FileInfo file_info;
                 string file_name = "";
                 while ((file_info = enumerator.next_file ()) != null) {
-                    file_name = file_info.get_name ();
+                    if (file_info.get_file_type () == FileType.REGULAR) {
+                        file_name = file_info.get_name ();
+                        break;
+                    }
                 }
 
                 path = Path.build_filename (path, file_name);
--- a/src/LoginOptions/UserLogin.vala
+++ b/src/LoginOptions/UserLogin.vala
@@ -21,10 +21,29 @@
 public class UserLogin : LoginOption {
 
     public LightDM.User lightdm_user { get; private set; }
+    private string background_path;
 
     public UserLogin (int index, LightDM.User user) {
         base (index);
         this.lightdm_user = user;
+
+        try {
+            string path = Path.build_filename ("/var", "lib", "lightdm-data", lightdm_user.name, "wallpaper");
+            var background_directory = File.new_for_path (path);
+            var enumerator = background_directory.enumerate_children (FileAttribute.STANDARD_NAME, 0);
+            
+            FileInfo file_info;
+            string file_name = "";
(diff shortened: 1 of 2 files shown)


src/LoginOptions/UserLogin.vala
```diff
@@ -34,7 +34,28 @@ public class UserLogin : LoginOption {
...
+                        file_name = file_info.get_name ();
+                    }
+
+                    path = Path.build_filename (path, file_name);
+                    return path;
+                } catch (Error e) {
+                    warning (e.message);
+                    return "";
+                }
+
```
donadigo: Remove this newline.

src/LoginOptions/UserLogin.vala
```diff
@@ -34,7 +34,28 @@ public class UserLogin : LoginOption {
     }
 
     public override string background {
-        get {
+        owned get {
+            if (lightdm_user.background == null) {
+                try {
+                    string path = Path.build_filename ("/var", "lib", "lightdm-data", lightdm_user.name, "wallpaper/");
```
donadigo: Do not need the last `/` here. Just "wallpaper".

src/LoginOptions/UserLogin.vala
```diff
@@ -35,6 +54,10 @@ public class UserLogin : LoginOption {
 
     public override string background {
         get {
+            if (lightdm_user.background == null) {
```
donadigo: Move that also to the constructor, this should always only return the background_path.

src/LoginOptions/UserLogin.vala
```diff
@@ -21,10 +21,33 @@
...
+        
+        if (lightdm_user.background == null) {
+            try {
+                string path = Path.build_filename ("/var", "lib", "lightdm-data", lightdm_user.name, "wallpaper");
+                var background_directory = File.new_for_path (path);
+                var enumerator = background_directory.enumerate_children (FileAttribute.STANDARD_NAME, 0);
+                
+                FileInfo file_info;
+                string file_name = "";
+                while ((file_info = enumerator.next_file ()) != null) {
```
donadigo: I think we should just pick the first normal file here to prevent picking directories, so if you are iterating over files in this directory you could do:
```
if (file_info.get_file_type () == FileType.REGULAR) {
    file_name = file_info.get_name ();
    break;
}
```

mdh34: done!

Djaler:
This works for me, thank you

Djaler:
@danrabbit can we marge this?

danrabbit:
Tested and I can confirm this fixes the regression

PR #97, opened 2018-03-31
RDNN

mdh34:
change naming from `pantheon-greeter` to `io.elementary.greeter`, as per RDNN spec


danrabbit:
Adding Cody for review here because this likely requires changes to configuration files

PR #109, opened 2018-05-15
Revert package rename

codygarver:
I want the package rename handled in a follow up PR since it's lower priority but still needs to be handled properly with a working transitional package.


codygarver:
This also updates the Homepage URL in the control file.

Issue #132, opened 2018-07-06
Provide dbus interface to the panel


danrabbit:
We need to provide the interface for getting background variance to the panel so that it can be transparent and/or light/dark where appropriate

donadigo:
I think this would be resolved automatically now that we have a compositor in place that would draw the background and have the wingpanel gala plugin just loaded in compositor? This is how it's currently done in a normal session.

torikulhabib:
in my opinion, should be in greeter mode, the wingpanel is positioned above the greeter.
because when the wingpanel appears the greeter is like being pressed down.

if not mistaken wingpanel ever experienced a bug, where all the maximized applications are under the wingpanel so that the maximum button and the lid are closed by wingpanel.

maybe if an exception is done in greeter, wingpanel is above the greeter maybe
Dark of wingpanel fixed.
just opinion

danrabbit:
@donadigo the problem is that the compositor can't draw the background right now because Clutter can't have a transparent background

danrabbit:
I'm going to drop this from the beta2 milestone because this is now an enhancement and no longer a regression

danrabbit:
I'm going to close this since the new greeter design resolves this automatically

PR #154, opened 2018-09-08
fix wingpanel black on greeter

torikulhabib:
https://github.com/elementary/greeter/issues/132


vjr:
Thank you @torikulhabib for this nice fix!

Would the devs please merge this? I tested and it also resolves a wallpaper sizing glitch during login.

The wallpaper in greeter was scaled down below the wingpanel area so upon login it would scale back up to the full desktop area but this is now fixed.

PR #167, opened 2018-10-18
Meson: correct install_path

worldofpeace:
This has been the correct way since meson 0.37.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,9 @@
 project(
     'io.elementary.greeter',
     'vala', 'c',
-    version: '3.3.0')
+    version: '3.3.0',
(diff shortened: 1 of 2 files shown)


worldofpeace:
Yeah also just noticed that this isn't the only occurrence of this so I'll fix the rest of them.

worldofpeace:
@codygarver There's no `meson_version : ' '` in 
https://github.com/elementary/greeter/blob/25584773f84d10093defaa014883fba93f2d66eb/meson.build#L1

codygarver:
@worldofpeace then it should be added

worldofpeace:
:fireworks: :sparkles:

Issue #131, opened 2018-07-06
Suspending (graphical glitch + power button only)


eccathcart:
![screenshot from 2018-07-06 21 31 48](https://user-images.githubusercontent.com/7200057/42371787-94749596-8164-11e8-89fd-dcdfe94218e8.png)

When suspending the computer or letting it fall asleep, the only way I can awake it again is by holding the power button on my desktop. Pressing anything on the keyboard or clicking the mouse doesn't get it to wake.

Another issue is the graphical issues that appear. When awoken, the wallpaper on the greeter is severely distorted, looking similar to the background in the multitasking view (in above image). The only way to get rid of it is to restart my computer but it reoccurs any time I put my computer to sleep.

~ I'm currently running Juno 5.0 Beta 1. This never happened when I was using Loki. Sorry if I posted this in the wrong place.

mdh34:
Are you using NVIDIA drivers?

eccathcart:
Yeah, I'm using the latest one that eOS gives (390.48).

mdh34:
In that case this is a duplicate of #gala/208 as the greeter uses a stripped down version as a compositor

mdh34:
I'll close this so we can track this in one place

mdh34:
Oops, turns out we don't actually use gala but rather a basic mutter compositor, so this is a greeter issue

eccathcart:
With the recent updates, the wallpaper issue has seem to gone away. I'm still getting the issue where I can only wake my computer by pressing the power button though.

peteruithoven:
Maybe clarify you initial comment or open a separate one? It's inconvenient to report on 2 issues in one issue.

jlnr:
All the screenshots so far have shown gala, not greeter. Here's a photo of greeter specifically:

![img_3661](https://user-images.githubusercontent.com/70772/47261375-15451300-d4ce-11e8-8d05-c4e76f339f67.jpg)

cassidyjames:
Marking as high priority since this affects a lot of people and we continue to get reports of it here and via social media.

kbknapp:
Same thing for me, only way to wake from sleep is power button, then get the "neon checkerboard" for a second or two as it wakes. Not using NVidia though. Here's my specs:

```
OS: elementary OS 5.0 Juno x86_64 
Host: Precision 7710 
Kernel: 4.20.3-042003-generic 
Uptime: 4 days, 3 hours, 32 mins 
Packages: 2140 
[...]

PR #170, opened 2018-10-20
Nvidia drivers: Refresh background after suspend (#131)

jlnr:
Exactly the same fix as in https://github.com/elementary/gala/pull/330.

This should fix the first half of #131, but I don't have my Nvidia machine with me, so I can't test this myself right now.

### DIFF
--- a/compositor/SystemBackground.vala
+++ b/compositor/SystemBackground.vala
@@ -21,13 +21,13 @@
 
 public class Greeter.SystemBackground : Meta.BackgroundActor {
     const Clutter.Color DEFAULT_BACKGROUND_COLOR = { 0x2e, 0x34, 0x36, 0xff };
-    
-    static Meta.Background? system_background = null;
 
-    public signal void loaded ();
+    static Meta.Background? system_background = null;
 
     public SystemBackground (Meta.Screen screen) {
         Object (meta_screen: screen, monitor: 0);
+
+        background = system_background;
     }
 
     construct {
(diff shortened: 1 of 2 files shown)


jlnr:
@leonardo-lemos Thanks for testing this PR! It turns out the greeter is not using `compositor/Wallpaper.vala` (like gala) but `src/Widgets/Wallpaper.vala` (for the crossfade effect). Consequently, calling `Meta.Background.refresh_all ()` did not fix the issue.

I've added code to have `PantheonGreeter` wait for resume events on Nvidia systems, but since `PantheonGreeter` does not have access to Mutter(?) (for the `gl_video_memory_purged` signal), I've had to use the OpenGL-based workaround that I just deleted in the compositor. Meh.

When a resume event on an Nvidia-based system is detected, I now call `wallpaper.reposition ()`, which seems to restore the wallpaper after resuming. However, the user list and clock are still garbled. I'm not sure why these use Clutter instead of Gtk, but no combination of `ensure_redraw` and `queue_redraw` seems to ungarble these widgets. They keep using purged textures.

One thing that almost works is to add `remove (clutter); add (clutter);` in `prepare_for_sleep`. But *sometimes* I ended up with a screen where both the users *and* the backgrounds where broken, so I've taken it out, even though it worked in 2 out of 3 attempts.

I don't have any more time to spend on this for this month. Maybe rewriting the user list & clock in pure Gtk would be easier? This, along with the changes in this PR, might be enough to fix the issue.

jlnr:
@leonardo-lemos Thanks for the pointer. I won't stop anyone from finishing this PR, but I don't think I'll invest any more time here until #181 is merged.

Forgot to attach a photo I took of the current state of things – background works, user list is garbled.

img_3794

jlnr:
I've removed my changes to the parts that have been rewritten in GTK. It's not a very simple fix, and should stop the dark grey background from corruption after resume (successfully tested here).

jlnr:
I didn't have a lot of time for testing today, but I haven't seen any corruption on the master branch when resuming at the greeter. So I'm not sure if there's an actual issue left to fix here.

However, I have rebased this branch, and it seems to still be free of corruption, but with two changes:

* I've replaced the logind workaround with the more specific `gl_video_memory_purged` signal which was built precisely to recover OpenGL textures.
* I have made the `refresh` method on `Greeter.SystemBackground` static so that it can be called without creating a temporary dummy instance (which only refreshes the single, static `system_background` anyway).

lvitals:
Okay, I did the tests, gl_video_memory_purged retrieves the OpenGL textures, and then the static update method is called to redo the color and the background image.

leonardo-lemos:
I'm closing this PR since #206 and #208 supersede this.

PR #177, opened 2018-10-26
Spelling: Occurred

comradekingu:
https://en.wiktionary.org/wiki/occurred
Spotted by @Sporiff on Weblate


cassidyjames:
This looks like it needs to match an error coming from fprintd, so if there's a typo there, we'd have to match it.

comradekingu:
@cassidyjames As per https://github.com/dsd/fprintd/pull/4/files ?

cassidyjames:
@comradekingu indeed!

leonardo-lemos:
Blocked until fprintd/pull/4 is merged.

tintou:
Note that the string change has been merged a while ago in https://gitlab.freedesktop.org/libfprint/fprintd/commit/51f4dce4e778bcce18659aa1d8a22a4d1cfdbd1f (the right repository for fprintd)

PR #185, opened 2018-11-12
Always show usernames

### LINKED ISSUE #69: No user names visible at Pantheon Greeter until click
After updating to loki, at the login screen all users  are represented by blue icons. I don't see the login names behind the blue  icons until I click in the space behind the icon. (I have to click icon  by icon in order to find the correct user to login.) 
The name should be shown at the side of the icon, just like it used to be on elementary 0.3.
This is really bad for the usability on multi user systems.

I found at least two threads talking about the problem:
https://elementaryos.stackexchange.com/questions/8220/no-user-names-visible-at-pantheon-greeter-until-click
https://elementaryos.stackexchange.com/questions/11411/show-user-names-on-the-login-screen

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/49044303-no-user-names-visible-at-pantheon-greeter-until-click?utm_campaign=plugin&utm_content=tracker%2F60068889&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60068889&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #69

### DIFF
--- a/src/Cards/BackgroundImage.vala
+++ b/src/Cards/BackgroundImage.vala
@@ -1,16 +1,10 @@
 public class Greeter.BackgroundImage : Gtk.EventBox {
-    public bool round_bottom { get; set; default=false; }
-
     private uint last_size_hash = 0;
     private Gdk.Pixbuf full_pixbuf;
     private Gdk.Pixbuf fitting_pixbuf;
 
     construct {
         height_request = 150;
-
-        notify["round-bottom"].connect (() => {
-            queue_draw ();
-        });
     }
 
     public BackgroundImage (string? path) {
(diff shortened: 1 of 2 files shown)


cassidyjames:
The design discussion from Slack can basically be summed up: Harvey and I like the cleaner design with wallpaper + avatar as the user's identity, and the issue could be solved instead by ensuring fallback avatars (or default avatars set via Initial Setup/User Accounts settings) are unique, like the [Material Design avatars](https://github.com/lincanbin/Material-Design-Avatars). Unless/until there's work toward that, we could show the name all the time.

PR #186, opened 2018-11-12
UserCard: Add collapsed class to unfocused cards

danrabbit:


src/Cards/UserCard.vala
```diff
@@ -10,6 +10,7 @@ public class Greeter.UserCard : Gtk.Revealer {
     public double reveal_ratio { get; private set; default = 0.0; }
 
     private Gtk.Revealer form_revealer;
+    private Gtk.StyleContext main_grid_style_context;
```
tintou:
```suggestion
    private weak Gtk.StyleContext main_grid_style_context;
```

PR #189, opened 2018-11-13
 UserCard: use wallpaper path fallback code from old greeter

danrabbit:
https://github.com/elementary/greeter/blob/767bbcdce0f3a3189003ed2a6b4b2a6b16b2c9ba/src/LoginOptions/UserLogin.vala#L30

### DIFF
--- a/src/Cards/UserCard.vala
+++ b/src/Cards/UserCard.vala
@@ -64,31 +64,24 @@ public class Greeter.UserCard : Gtk.Revealer {
         form_revealer.transition_type = Gtk.RevealerTransitionType.SLIDE_DOWN;
         form_revealer.add (form_grid);
 
-        string? background_path;
+        var background_path = lightdm_user.background;
 
-        if (lightdm_user.background == null) {
+        if (background_path == null) {
             try {
-                string path = Path.build_filename ("/", "var", "lib", "lightdm-data", lightdm_user.name, "wallpaper");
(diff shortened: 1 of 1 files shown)


src/Cards/UserCard.vala
```diff
@@ -64,7 +64,31 @@ public class Greeter.UserCard : Gtk.Revealer {
...
+            try {
+                string path = Path.build_filename ("/", "var", "lib", "lightdm-data", lightdm_user.name, "wallpaper");
+                var background_directory = File.new_for_path (path);
+                var enumerator = background_directory.enumerate_children (FileAttribute.STANDARD_NAME, 0);
+
+                FileInfo file_info;
+                string file_name = "";
+                while ((file_info = enumerator.next_file ()) != null) {
+                    if (file_info.get_file_type () == FileType.REGULAR) {
+                        file_name = file_info.get_name ();
```
tintou:
```suggestion
                        background_path = Path.build_filename (path, file_info.get_name ());
```

src/Cards/UserCard.vala
```diff
@@ -64,7 +64,31 @@ public class Greeter.UserCard : Gtk.Revealer {
...
+                        break;
+                    }
+                }
+
+                path = Path.build_filename (path, file_name);
```
tintou: remove the two lines here, we got the background_path variable in the loop directly and the BackgroundImage can handle `null` background_path

PR #190, opened 2018-11-13
Shake the whole card on login error

danrabbit:
Instead of just shaking the password entry, shake the whole card


src/Cards/UserCard.vala
```diff
@@ -208,6 +208,12 @@ public class Greeter.UserCard : Greeter.BaseCard {
     }
 
     public override void wrong_credentials () {
-        password_entry.animate_error ();
+        main_grid_style_context.add_class (Gtk.STYLE_CLASS_ERROR);
```
tintou: You probably only want to add the error class to the password entry

src/Cards/ManualCard.vala
```diff
@@ -53,6 +54,13 @@ public class Greeter.ManualCard : Greeter.BaseCard {
     }
 
     public override void wrong_credentials () {
-        password_entry.animate_error ();
+        weak Gtk.StyleContext style_context = main_grid.get_style_context ();
+        style_context.add_class (Gtk.STYLE_CLASS_ERROR);
```
tintou: You probably only want to add the error class to the username/password entries

PR #199, opened 2018-11-15
Update it.po

maerlin:
Removed " %p" from AM/PM showing since it's not used in Italian. It is used always either as a 24h format (23.50) or 12h format but without AM/PM (11.50).


tintou:
Hi! We are using Weblate to translate everything in elementary (see how to participate https://elementary.io/it/get-involved#translations ). You can submit the translation using this platform, the Italian translation of this string is here: https://weblate.elementary.io/translate/desktop/greeter/it/?checksum=9e89ac7515032368

Issue #213, opened 2018-11-26
Hitting Super+T opens Terminal


cassidyjames:
It seems like it backgrounds right away since the greeter window is always on top, but that should really not be possible to open, as it could lead to exploits.

PR #218, opened 2018-11-27
Simple blacklist media keys

torikulhabib:
problem difference user and lightdm user.
Simple blacklist media keys for terminal without change code.
and remove this after found alternatif for lightdm.
https://github.com/elementary/greeter/issues/213


peteruithoven:
This feels a bit leaky? What if someone or a nefarious script adds a custom keyboard combination? Wouldn't a whitelist be more appropriate?
How do other greeters handle this?

tintou:
I'm pretty sure that the media keys are handled by the compositor itself, so it's probably safer and easier to filter them from there

PR #240, opened 2019-03-06
MainWindow: Hard code stylesheet and icons

### LINKED ISSUE #239: hard-code GTK and icon theme used by the greeter
On systems where the default GTK and icon theme can't be overridden system-wide (e.g. on fedora, and most non-elementaryOS distributions), the greeter has visual issues.

I think in this case, where the program really is explicitly dependent on specific GTK styles from the elementary GTK stylesheet and on specific icons from the elementary icon theme, it would be appropriate to disregard the system settings and hard-code the themes instead, to fix potential issues.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/70950892-hard-code-gtk-and-icon-theme-used-by-the-greeter?utm_campaign=plugin&utm_content=tracker%2F60068889&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60068889&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #239

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -60,12 +60,12 @@ public class Greeter.MainWindow : Gtk.ApplicationWindow {
 
         try {
             var settings = Gtk.Settings.get_default ();
+            settings.gtk_icon_theme_name = "elementary";
+            settings.gtk_theme_name = "elementary";
+
             var css_provider = Gtk.CssProvider.get_named (settings.gtk_theme_name, "dark");
             guest_login_button.get_style_context ().add_provider (css_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
             manual_login_button.get_style_context ().add_provider (css_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
-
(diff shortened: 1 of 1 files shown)


src/MainWindow.vala
```diff
@@ -63,6 +63,9 @@ public class Greeter.MainWindow : Gtk.ApplicationWindow {
             var css_provider = Gtk.CssProvider.get_named (settings.gtk_theme_name, "dark");
             guest_login_button.get_style_context ().add_provider (css_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
             manual_login_button.get_style_context ().add_provider (css_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
+
+            settings.gtk_icon_theme_name = "elementary";
+            settings.gtk_theme_name = "elementary";
```
jeremypw: Is it correct to change this after using the original in line 63?

danrabbit: That's a good point :grimacing:

Issue #245, opened 2019-03-15
Resize jump


peteruithoven:
I've got a HiDPI screen and I noticed that in a lot of cases the Greeter seems to start out smaller and then resizes to a normal size:
![VID_20190315_201448~2 mp4](https://user-images.githubusercontent.com/523210/54457371-7ad9ba80-4761-11e9-9896-d13f7de9fb6b.gif)

It might be related to: #144 as when I see this jump happen the field won't be focused. Even after: https://github.com/elementary/greeter/pull/243

danrabbit:
I think this is because we render the greeter before GSD is loaded. I wonder if it would be better to try to remember the last setting and load that in a way that is faster than waiting for GSD or if we should wait to render the greeter until after GSD is loaded

PR #243, opened 2019-03-15
Present main window and focus current card after users loaded.

jeremypw:
Fixes #144  (or improves it).

Difficult to be sure whether this completely fixes the problem on all systems under all circumstances, but it appears to do so on mine.

Issue #144, opened 2018-07-14
Greeter window sometimes doesn't grab focus [$60]


Djaler:
After last greeter updates, I noticed that when I boot into elementary, password input is not in focus

darthmaulbr:
I have the same issue.

jiteshsingh:
same issue but only when login screen appears first time after bootup

danrabbit:
Please do not comment "I have the same issue". Use the :+1: reaction

ratza:
I can confirm that this does not seem to be an issue with the latest version (3.2.0) on boot in Juno. It's still an issue when locking the screen. I didn't even realize how used I got to the fact that on Loki, when locking the session, even though the screen stays off the input is in focus - it allowed me to login without even waiting for the monitor to detect and display the screen.

peteruithoven:
I can only consistently reproduce this issue after Shutdown or Restart. The password field has focus for me after a Lock or Suspend. 

Is there a way for me to debug this? Maybe add some logging to journal?

yusren:
The worst thing on Juno's Login screen is that the keyboard doesn't work until we click on the screen.

cassidyjames:
Confirming and marking as high priority since this affects many users and we're getting a lot of mentions of it on social media.

tugcanolgun:
I can confirm this. This happens in my laptop too.

infuerano:
It is a problem which is annoying.

cassidyjames:
Please do not comment things like, "I have the same issue". Use the :+1: reaction to mark it as affecting you, and/or use the "Subscribe" button to get updates.

kt1024:
maybe this is relevant to #131 , here is what I found:
- on a notebook with single/internal intel video card, the password input box is auto focused. 
- on notebook with 2 video card (intel + nvidia), input box is not focused, and 'unknown display' is always activated. 
above test is done with fresh juno install.

seesemichaelj:
I am able to reproduce this issue consistently on boot and from locking.

I have also found a hotfix that really doesn't address the root problem. At the end of the constructor in [PantheonGreeter.vala](https://github.com/elementary/greeter/blob/master/src/PantheonGreeter.vala#L211), if I add the following code, I can get it to focus the login box automatically after some delay:

```
Timeout.add (5000, () => {
    userlist_actor.get_current_loginbox ().pass_focus ();
    this.get_window ().focus (Gdk.CURRENT_TIME);
    return false;
});
```

[...]

PR #246, opened 2019-03-16
Re-grab focus after dpi changed

peteruithoven:
I've noticed this resize jump #245 and long thought that might be a cause of losing focus: #144. 
This uses the same principle as #243 to re-grab focus after the dpi changed.
This seems to consistently fix it for me.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -239,6 +239,16 @@ public class Greeter.MainWindow : Gtk.ApplicationWindow {
             return false;
         });
 
+        // regrab focus when dpi changed
+        get_screen ().monitors_changed.connect(() => {
+            present ();
+
+            /* Ensure current card is focused if set */
+            if (current_card != null) {
+                current_card.grab_focus ();
+            }
+        });
+
         destroy.connect (() => {
             Gtk.main_quit ();
         });


lhorace:
hi @peteruithoven , what is the easiest way to test this to see it works?

peteruithoven:
@lhorace Try the `master` branch. Notice that the password field isn't focused. 
Then try this branch, the field should be focused.

lhorace:
How? When you say master branch, disrupt my desktop? Compile on elmentaryOS?

lhorace:
It be allot easier if you explain ?

peteruithoven:
Sorry, you didn't really give an indication of your experience level. Explaining everything in an issue is easier for you but not very time effective or scalable. Have you tried following the readme of the project? 
https://github.com/elementary/greeter#building-and-installation

There are also plenty of resources available on using Git, which should explain what things like branches are.

If you don't have experience with these matters the Greeter might not be the easiest project to start with though. I'd recommend first trying out apps like Calendar or indicators or the switchboard plugs. 
Some general resources: 
https://elementary.io/docs/code/os-dev
https://elementary.io/docs/code/getting-started#getting-started
https://elementary.io/get-involved#desktop-development

peteruithoven:
This fix was included in #247 which has been merged.

PR #251, opened 2019-03-21
SettingsDaemon: Disable GSD media shortcuts for apps

### LINKED ISSUE #213: Hitting Super+T opens Terminal

danrabbit:
Fixes #213


peteruithoven:
This feels a bit leaky? What if someone or a nefarious script adds a custom keyboard combination? Wouldn't a whitelist be more appropriate?
How do other greeters handle this?

tintou:
Yeah, and if you look at all the keys that the daemon provides, there are way too many keys for the greeter

danrabbit:
I’m not sure how else to handle it since we want the media keys plugin to handle things like screen brightness

cassidyjames:
And volume, likely, so you could easily mute before unlocking.

tintou:
Please test #255 as it should be the proper solution

PR #255, opened 2019-03-28
Media Keys: Do not send unwanted signals

### LINKED ISSUE #213: Hitting Super+T opens Terminal

tintou:
Fixes #213

### DIFF
--- a/compositor/DBusAccelerator.vala
+++ b/compositor/DBusAccelerator.vala
@@ -17,9 +17,40 @@
  */
 
 namespace GreeterCompositor {
+    /**
+     * ActionMode:
+     * @NONE: block action
+     * @NORMAL: allow action when in window mode, e.g. when the focus is in an application window
+     * @OVERVIEW: allow action while the overview is active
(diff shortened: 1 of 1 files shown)


peteruithoven:
This uses the flags as defined here: 
https://gitlab.gnome.org/GNOME/gnome-settings-daemon/blob/master/plugins/media-keys/shell-action-modes.h
and configured per key here: 
https://gitlab.gnome.org/GNOME/gnome-settings-daemon/blob/master/plugins/media-keys/shortcuts-list.h

PR #258, opened 2019-03-28
UserCard: Make session button insensitive when user is logged in

### LINKED ISSUE #237: don't show session chooser widget on the lock screen
Right now, the session chooser widget is always shown - even on the "lock screen", where choosing a different session doesn't make any sense.

(This is true for both the latest stable release and git master.)

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/70817627-don-t-show-session-chooser-widget-on-the-lock-screen?utm_campaign=plugin&utm_content=tracker%2F60068889&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60068889&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #237

### DIFF
--- a/src/Cards/UserCard.vala
+++ b/src/Cards/UserCard.vala
@@ -139,6 +139,9 @@ public class Greeter.UserCard : Greeter.BaseCard {
             logged_in.halign = logged_in.valign = Gtk.Align.END;
 
             avatar_overlay.add_overlay (logged_in);
+
+            session_button.sensitive = false;
+            session_button.tooltip_text = (_("Session type cannot be changed while logged in"));
         }
 
         var card_overlay = new Gtk.Overlay ();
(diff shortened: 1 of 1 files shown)


danrabbit:
I'd like feedback from @cassidyjames on tooltip copy

src/Cards/UserCard.vala
```diff
@@ -139,6 +139,9 @@ public class Greeter.UserCard : Greeter.BaseCard {
             logged_in.halign = logged_in.valign = Gtk.Align.END;
 
             avatar_overlay.add_overlay (logged_in);
+
+            session_button.sensitive = false;
+            session_button.tooltip_text = (_("Session type cannot be changed while logged in"));
```
cassidyjames: @danrabbit what about:

```suggestion
            session_button.tooltip_text = (_("Session cannot be changed while user is logged in"));
```

PR #259, opened 2019-03-29
UserCard: Check with accountsservice if user is locked

### LINKED ISSUE #236: Hide disabled user accounts
If a user account is disabled in System Settings → User Accounts, it should not show up on the greeter.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/70792391-hide-disabled-user-accounts?utm_campaign=plugin&utm_content=tracker%2F60068889&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60068889&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #236 

Screenshot from 2019-03-28 23 16 04@2x


src/Cards/UserCard.vala
```diff
@@ -67,9 +67,17 @@ public class Greeter.UserCard : Greeter.BaseCard {
         this.bind_property ("connecting", login_button, "sensitive", GLib.BindingFlags.INVERT_BOOLEAN);
         login_button.get_style_context ().add_class (Gtk.STYLE_CLASS_SUGGESTED_ACTION);
 
+        var disabled_message = new Gtk.Label (_("Account disabled by an administrator"));
+        disabled_message.margin_top = 3;
+        disabled_message.max_width_chars = 30;
+        disabled_message.valign = Gtk.Align.START;
+        disabled_message.wrap = true;
+        disabled_message.get_style_context ().add_class (Gtk.STYLE_CLASS_ERROR);
```
cassidyjames: This feels a little heavy-handed/loud, don't you think? It's not like some critical error, it's just locked. :smile: 

```suggestion
```

src/Cards/UserCard.vala
```diff
@@ -67,9 +67,17 @@ public class Greeter.UserCard : Greeter.BaseCard {
         this.bind_property ("connecting", login_button, "sensitive", GLib.BindingFlags.INVERT_BOOLEAN);
         login_button.get_style_context ().add_class (Gtk.STYLE_CLASS_SUGGESTED_ACTION);
 
+        var disabled_message = new Gtk.Label (_("Account disabled by an administrator"));
```
cassidyjames: Do we know it was disabled by an administrator? Is it just as useful to say "Account disabled"?

decathorpe:
I seem to remember that this "locked account" functionality is is only added to accountsservice with ubuntu-specific patches. In this case this should probably be behind conditional compilation.

But don't take my word for it, I can't check right now since I'm not at my desk

tintou:
@decathorpe Looks like it's in master so it should be in Fedora https://gitlab.freedesktop.org/accountsservice/accountsservice/blob/master/data/org.freedesktop.Accounts.User.xml#L851

src/Cards/UserCard.vala
```diff
@@ -151,6 +164,15 @@ public class Greeter.UserCard : Greeter.BaseCard {
 
         add (card_overlay);
 
+        var user = Act.UserManager.get_default ().get_user (lightdm_user.name);
```
tintou: This line is not needed, there is already a `lightdm_user` property

src/Cards/UserCard.vala
```diff
@@ -151,6 +164,15 @@ public class Greeter.UserCard : Greeter.BaseCard {
...
+        user.notify["is-loaded"].connect (() => {
+            if (user.locked) {
```
tintou: better create a private method for this and calling it once first as we are not sure that the `is-loaded` property will ever change, also add the `locked = false` counterpart

tintou:
PR opened here https://github.com/CanonicalLtd/lightdm/pull/66

src/Cards/UserCard.vala
danrabbit: I created property bindings for the two booleans and a private method for the stack switcher. We should now handle if the user is unlocked while the greeter is up.

PR #260, opened 2019-03-30
Fade greeter in to avoid jank

cassidyjames:
This waits until we do the maximize-and-focus to also fade the greeter in, avoiding any rendering weirdness that happens earlier (like seeing it scaled to 1x on a 2x display).

I have two issues with this:

- [ ] It works fine when running in-session, but seemingly animates twice in the greeter session.
- [ ] It feels weird to be manually animating… would it be better to use a CSS class for this?

### DIFF
--- a/data/MainWindow.css
+++ b/data/MainWindow.css
@@ -1,0 +1,9 @@
+.background {
+    opacity: 0;
+    transition: opacity 1s ease;
+}
+
+.background.initialized {
+    opacity: 1;
+}
+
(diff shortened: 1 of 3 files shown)


peteruithoven:
>  but seemingly animates twice in the greeter session.

Was this solved?

peteruithoven:
Great improvement! 
I did notice that the wingpanel still jumps though: #264

cassidyjames:
@peteruithoven ah yeah the double flash was fixed by moving to CSS. WingPanel is actually entirely separate from the Greeter window, so we can't control it from here.

peteruithoven:
I was afraid of that, luckily it's a separate issue for now. We can always look into it after 5.1. Although...
Although... we do [start the wingpanel from the Greeter](https://github.com/elementary/greeter/blob/master/src/Application.vala#L47). Maybe we can start it at the same moment you add the `initialized`. (With a check that it can only be started once)

Issue #275, opened 2019-05-08
Greeter does not respect time format settings


MadByteDE:
The problem with the old greeter version was that the 24h format had AM/PM labels and in the german translation also was misaligned because of bad/too long translations.

The new version now completely ignores system preferences and displays a 12h format time.

![comparison](https://user-images.githubusercontent.com/4752038/57390645-24b35180-71bd-11e9-8ed4-f52b7f7edb64.jpg)

OS: Elementary OS 5.0
Screen Res: 1920x1080
Greeter version: Latest master (8th May 2019).

peteruithoven:
There is a more general issue that the Greeter doesn't follow settings: https://github.com/elementary/greeter/issues?utf8=%E2%9C%93&q=user+settings+is%3Aopen+is%3Aissue+
It's challenging because there can be multiple users each having their own settings.

danrabbit:
This is fixed within the scope of the greeter repo. There is another branch out against the Date & Time settings plugin for Switchboard that will need to be merged for the issue to be resolved completely, but as far as Greeter is concerned the issue is fixed here.

PR #279, opened 2019-05-21
Use the current selected user time format

tintou:
We then need to replace the GSettings with this DBus parameter that is now exposed

### DIFF
--- a/data/io.elementary.greeter.AccountsService.policy
+++ b/data/io.elementary.greeter.AccountsService.policy
@@ -1,32 +1,32 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <policyconfig>
-        <action id="io.elementary.greeter.AccountsService.ModifyOwn">
-                <description gettext-domain="io.elementary.greeter">Set properties of own user</description>
-                <message gettext-domain="io.elementary.greeter">Authentication is required to set one's own greeter properties.</message>
-                <defaults>
-                        <allow_any>yes</allow_any>
-                        <allow_inactive>yes</allow_inactive>
-                        <allow_active>yes</allow_active>
-                </defaults>
-        </action>
+  <action id="io.elementary.greeter.AccountsService.ModifyOwn">
+    <description gettext-domain="io.elementary.greeter">Set properties of own user</description>
+    <message gettext-domain="io.elementary.greeter">Authentication is required to set one's own greeter properties.</message>
+    <defaults>
+      <allow_any>yes</allow_any>
+      <allow_inactive>yes</allow_inactive>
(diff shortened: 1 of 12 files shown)


peteruithoven:
This is great! This should fix quite some issues. 
What's the relationship between the `io.elementary.greeter.AccountsService` and the `org.freedesktop.Accounts`? 
This accounts service could it also store & share things like the display, power, etc settings?

peteruithoven:
This should fix: #275

data/io.elementary.greeter.AccountsService.policy
```diff
@@ -0,0 +1,32 @@
...
+                <message gettext-domain="io.elementary.greeter">Authentication is required to set one's own greeter properties.</message>
+                <defaults>
+                        <allow_any>yes</allow_any>
+                        <allow_inactive>yes</allow_inactive>
+                        <allow_active>yes</allow_active>
+                </defaults>
+        </action>
+
+        <action id="io.elementary.greeter.AccountsService.ReadAny">
+                <description gettext-domain="io.elementary.greeter">Set properties of any user</description>
```
vjr: This should say `Get properties...` instead of `Set properties...` ?

tintou:
> This is great! This should fix quite some issues.
> What's the relationship between the `io.elementary.greeter.AccountsService` and the `org.freedesktop.Accounts`?
> This accounts service could it also store & share things like the display, power, etc settings?
> 
> I found some info on this AccountsService: https://www.freedesktop.org/wiki/Software/AccountsService/
> It's a pity they've not made the actual docs easily accessible.

AccountsService has now an extension system, it is not really well documented but there is at least one example here https://gitlab.freedesktop.org/accountsservice/accountsservice/blob/master/doc/org.mydm.MyDisplayManager.Wallpaper.xml
The DBus interface is then "mounted" by the accountsservice daemon which is available to the whole system.

data/io.elementary.greeter.AccountsService.policy
tintou: Thanks for noticing, it's now fixed

decathorpe:
So, are these new datetime settings only going to work if the session was started from the elementary greeter? I assume GDM et al. don't support this accountsservice extension (yet).

tintou:
As long as the AccountsService files are installed it should work. GDM won't use it of course as they are not interested in using the user settings there

decathorpe:
That's great news :) Thanks for the clarification.

getsnoopy:
When can this change be expected to go into a release?

cassidyjames:
It's on the todo list for the next release, so any day now. :smile:

PR #284, opened 2019-05-28
Focus password after keyboard navigation

### LINKED ISSUE #280: Focus password entry after every keyboard or mouse navigation
The password field is only focused when the greeter starts up and initial user is selected. If you switch users using keyboard left or right keys then the next user's password field is not auto selected, which it should be for convenience.

Similarly if mouse navigation (scroll up/down) is added then here too the password field should get auto focused after moving to the next user.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/74696619-focus-password-entry-after-every-keyboard-or-mouse-navigation?utm_campaign=plugin&utm_content=tracker%2F60068889&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60068889&utm_medium=issues&utm_source=github).
</bountysource-plugin>

vjr:
Fixes #280


danrabbit:
Even though the entry isn't focused after navigating, typing seems to re-focus it right away as far as I can tell. Is that not the case for you?

If that is the case, I'm inclined to close the linked issue report. Thoughts from @elementary/ux ?

vjr:
In my case it is/was not clear to me that the password could be typed anyway because there is no flashing cursor in the password input - causing me to use the mouse to enable password input - so my vote (if it makes any difference) is to merge this PR to have it more discoverable.

cassidyjames:
@danrabbit yeah it technically works but I get what @vjr is saying; it's not obvious that the entry is focused and that typing will focus it, so I think actually setting focus wouldn't be a bad idea. Signifying affordances and all that :wink:

isantop:
My thought is that it's probably a good idea to keep the placeholder text even when it's not normally going to be visible.

Issue #233, opened 2019-02-16
Two-finger horizontal swipe to switch acounts


cassidyjames:
Apparently I just expected this to work similarly to how workspaces work in the Multitasking View. I think it'd be handy to horizontal swipe over to focus other users on the greeter.

cassidyjames:
This was fixed in #285

PR #285, opened 2019-05-28
Add mouse scroll navigation

### LINKED ISSUE #281: Add mouse scroll up/down navigation to move to next user
Similar to what was on the "old" greeter, even though the user list moves left/right, the mouse scroll up or down should also move to the next user.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/74696696-add-mouse-scroll-up-down-navigation-to-move-to-next-user?utm_campaign=plugin&utm_content=tracker%2F60068889&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60068889&utm_medium=issues&utm_source=github).
</bountysource-plugin>

vjr:
Fixes #281

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -260,6 +260,12 @@ public class Greeter.MainWindow : Gtk.ApplicationWindow {
             case Gdk.ScrollDirection.DOWN:
                 activate_action ("next", null);
                 break;
+            case Gdk.ScrollDirection.LEFT:
+                activate_action ("previous", null);
+                break;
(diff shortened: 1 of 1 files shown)


vjr:
Also fixes #233

tintou:
Has it being tested with and without "natural scrolling" ?

danrabbit:
I think the greeter always uses natural scrolling since that's the default in the OS

Issue #302, opened 2019-06-29
Multiple monitors are recognized as one big monitor


veraguido:
In this case 2 monitors (different resolutions) are recognized as one very wide monitor, placing the login details vignettes and the clock between both of them
![IMG_20190629_092735](https://user-images.githubusercontent.com/5069483/60381688-9c308f00-9a50-11e9-84d5-7ebab1e6a857.jpg)

GPU: nvidia 1050ti (prime)

peteruithoven:
@robertgorbul noticed the same, details from https://github.com/elementary/greeter/issues/301:

**OS:** elementary OS 5.0 juno
**Kernel:** x86_64 Linux 4.18.0-25-generic
**Greeter:** pantheon-greeter_3.3.1+r773+pkg61_daily_ubuntu5.0.1_amd64
**Issue:**
When system starts up or locked with external monitor greeter is centered between two screens. External monitor is primary. Wingpanel not affected. After detach and reattach of external monitor greeter is displayed on internal monitor.
![IMG_20190629_103919](https://user-images.githubusercontent.com/9195693/60381255-54166a00-9a5a-11e9-8ab1-a6e6ac9cb465.jpg)
**P.S:** Sorry for the photo. I tried to get a screenshot but result wasn't appropriate unfortunately.

peteruithoven:
Are there any situations where you don't see this behavior? 
@robertgorbul are you also using a NVIDIA card?

I can't reproduce this issue: 
![IMG_20190701_205542-2000](https://user-images.githubusercontent.com/523210/60460385-e3cb3c80-9c43-11e9-8b82-67201f4e24aa.jpg)

danrabbit:
I'm also unable to reproduce this issue using a USB C output on my notebook for an external HDMI TV.

robertgorbul:
@peteruithoven it's always the same. 
My laptop have hybrid graphics - Intel HD4000 + AMD Radeon 8750M. 
Kernel 4.18.0-25-generic. Mesa 19.1.0 from Padoka stable.
Internal screen is 1366x768 as secondary and external Benq 2560x1440 as primary connected through HDMI. And just fyi I have a custom modeline:
```
  1 Section "Monitor"
  2   Identifier "HDMI-1"
  3   Modeline "2560x1440_55.00" 221.00  2560 2608 2640 2720  1440 1443 1447 1478  +HSync -VSync
  4   Option "PreferredMode" "2560x1440_55.00"
  5 EndSection
``` 
[...]

PR #307, opened 2019-07-01
Get primary monitor to fix multi-display size

danrabbit:
I can't reproduce #302 but maybe someone can test this potential fix?


peteruithoven:
One confirmed fix: https://github.com/elementary/greeter/issues/302#issuecomment-507811702

peteruithoven:
Second confirmed fix: https://github.com/elementary/greeter/issues/302#issuecomment-508987160

We should probably pull this out of draft mode?

PR #322, opened 2019-08-15
Show greeter on the monitor which has the mouse pointer

andreasfelix:
When connected to an external monitor, the greeter is sometimes shown on the laptop monitor which is inconvenient if the laptop is closed or half-closed. Therefore I think it would be useful to show the greeter always on the monitor which the mouse pointer is currently on.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -273,6 +273,11 @@ public class Greeter.MainWindow : Gtk.ApplicationWindow {
             maximize_and_focus ();
         });
 
+        leave_notify_event.connect (() => {
+            maximize_and_focus ();
+            return false;
+        });
+
         destroy.connect (() => {
             Gtk.main_quit ();
         });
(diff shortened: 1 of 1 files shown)


tintou:
Just of note for a future me: We want to move that logic in the Compositor at some point as it won't be possible to do this in Wayland

andreasfelix:
I just noticed that this doesn't work if one is moving the pointer from one monitor across the wingpanel to the other monitor. This happens because the `leave_notify_event` of the Greeter is triggered when entering the wingpanel whereby the pointer is still on the first monitor. This is especially a problem for a vertical multi-monitor setup with the primary monitor being at the bottom.

Instead of listing to `leave_notify_event` of the Greeter-window, it would be better to use an equivalent event for the monitors, but I cannot figure out how to do this.

cassidyjames:
@andreasfelix yeah, it's not perfect… It's still better than the previous state, though, imho. But yeah, we'll need a different solution for Wayland and to handle that panel edge case.

PR #335, opened 2019-08-29
MainWindow: Throw an error dialog on failed login

danrabbit:
And make sure this method always results in clearing the connecting state


Ericson2314:
It seems `e.message` isn't logged anywhere? I had a hard time debugging things and would prefer the  dialog box in lieu of proper logging.

PR #358, opened 2019-09-20
Blurred background

danirabbit:


cassidyjames:
This looks pretty, but I still have concerns about the spatial model and getting to something that makes more sense and is closer to Harvey's mockups. Blur makes a lot of sense for the background for logged in users if we do a separate login and lock screen like the mockups.

With this implementation, I'm getting _super_ bad performance and weird graphical glitches when switching between multiple users where the animations don't seem to complete, so widgets end up in the wrong places.

cassidyjames:
Video of the issues I'm seeing: https://youtu.be/j_VqmVQfXa4

webrunner1981:
I prefer like other people who are using Elementary OS every day the possibility to have a wallpaper instead this gray background. I made an investigation and it looks that there is a problem with the gala windows manager. I am unable to set the the texture.png as background, because I checked that the texture is hardcoded in this file SystemBackground.vala and dconf shows that io/elementary/desktop/gala/ does not exist. Seems that gala has been removed totally from the desktop. Why? I did not find any explanation.

cassidyjames:
@sysadmin-info the wallpaper is in the card for each user, which makes it easier to see multiple users at a glance and ensures the clock and other text remains readable (a common complaint of the old design). Gala has nothing to do with the greeter, and the texture is GResourced as a part of the greeter because it is part of the design.

sysadmin-info:
> @sysadmin-info the wallpaper is in the card for each user, which makes it easier to see multiple users at a glance and ensures the clock and other text remains readable (a common complaint of the old design). Gala has nothing to do with the greeter, and the texture is GResourced as a part of the greeter because it is part of the design.

Hello,

Final solution with the wallpaper for the lock screen:

Thread:
https://www.reddit.com/r/elementaryos/comments/cwisd3/is_there_something_wrong_with_the_new_pantheon/

Tested on Hera 5.1 

Downloaded this: https://github.com/elementary/greeter/archive/master.zip

from https://github.com/elementary/greeter/

Extracted the zip to /tmp

I opened terminal, sudo -i, typed root password. Then run gimp as root (just type gimp in terminal), saved jpg file as png in greeter-master/data folder (replaced previous texture.png file).

Edited the SystemBackground.vala file located in /tmp/greeter-master/compositor/ 

sudo vi /tmp/greeter-master/compositor/SystemBackground.vala

Changed path from:
var texture_file = GLib.File.new_for_uri ("resource:///io/elementary/desktop/gala/texture.png");

to:
var texture_file = GLib.File.new_for_uri ("resource:///io/elementary/greeter/texture.png");

Installed all dependencies listed here:

https://github.com/elementary/greeter/blob/master/README.md

gnome-settings-daemon >= 3.27
how to check version in terminal: 
sudo gnome-control-center --version

Installed dependencies

sudo apt install -y libaccountsservice-dev libgdk-pixbuf2.0-dev libgranite-dev libgtk-3-dev liblightdm-gobject-1-dev libwingpanel-2.0-dev libx11-dev meson valac libmutter-3 libmutter-3-dev

Compiled as root 
cd /tmp/greeter-master
meson build --prefix=/usr
cd build
ninja
ninja install

It is working. Finally! Hell yeah!

stsdc:
What if greeter background would be dynamically generated svg, based on a few main colours picked from a wallpaper?

sysadmin-info:
There is no need:
https://youtu.be/g-7aglPhf88
https://youtu.be/hAysr8Q32Ho
With Eric we have found a solution. 
Will developers include this I do not care. But hardcoding a background is not a very good solution. I hope at least you will let everyone to be able to modify this on their own according to those tutorials, because we just want to be able to modify this distro the way we want, not developers actually. I prefer my vision how should it looks like, so at least let us be able to modify this in each future version of Elementary OS by recompiling it.

LongJohn-Silver:
I am just coming here to say that I would really love if this would be the way eOS follows with the greeter, it looks much nicer than a grey background.

LongJohn-Silver:
Has been any decision yet on whether this will be done?

kenedyroger:
I think this feature is very cool, that gray background is boring, did you folks has been decided?

cassidyjames:
This draft PR has conflicts; someone will need to address those and the performance issues I pointed out before it could be considered.

Issue #352, opened 2019-09-06
Toggle password visibility


cassidyjames:
This has become pretty commonplace in password entries and can help it be more accessible to those who might have issues typing. I don't think this is a native GTK Entry feature, but it should be pretty easy to add.

cassidyjames:
Ah I guess the main design question here is: how? When creating a user, we use a checkbox below the entry. But that seems a bit heavy/ugly in this case. In other implementations, I've seen the right-side input icon used, but we're using that as the login button here.

4jNsY6fCVqZv:
Exciting, I support your idea! Before your issue, I had often wondered if this was intended as a Privacy by Design feature by elementary, not to offer the possibility to display the password/passphrase. From GNOME, however, I also know another way of dealing with it.

fitojb:
> Ah I guess the main design question here is: how?

In case it is helpful, [here’s Microsoft’s spec for this feature](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/master/PasswordReveal/explainer.md).

4jNsY6fCVqZv:
I just notice again how important it would be if I could have the password displayed for a short time. For security reasons, I use a password with many characters that exceed the normal length of the input field. If I mistype, which often happens, but the password field is not cleared directly, it doesn't help if the incorrect string is still displayed. I can't just find and correct the faulty parts. So I first have to clear the input field manually and then re-enter everything. This is a detour compared to the status before the last Greeter change.

capttrousers:
This is an important feature request. 

On my laptop, sometimes the number lock gets engaged automatically on boot up, so I type the wrong password just because the number pad is overlayed on the general keyboard. 

Gnome handles showing the password by right clicking the input field and there's an option in the context menu.  I think an even better design has the button on the input field directly, but Cassidy points out that the input for Greeter already has icons on the left and right. 

Can the toggle functionality be added "gnome" style in a right click-able context menu as a first step, and then the design of that password input field can be re-visited with this feature request in mind?  The icon on the left is just a key hole to indicate it's a password field, I dont know how helpful that is. And most people will typically just hit "Enter" after typing their password right, rather than using the button to submit?

4jNsY6fCVqZv:
@cassidyjames What do you think about starting with an item in the context menu first and going on from there?

cassidyjames:
[...]

PR #369, opened 2019-10-07
Clear password entry on error

Philip-Scott:
For a couple of times now, when I entered my password without really looking at my screen and pressing enter, for a split second I thought that Greeter was never in focus and that I didn't type anything. This was because the password was deleted on the `on_login ()` function. 

This PR just moves the clear password line from `on_login()` to `wrong_credentials()`, that way also when a user types the wrong password, the asterisks will stay in until the User Card starts shaking

### DIFF
--- a/src/Cards/ManualCard.vala
+++ b/src/Cards/ManualCard.vala
@@ -78,7 +78,6 @@ public class Greeter.ManualCard : Greeter.BaseCard {
     private void on_login () {
         connecting = true;
         do_connect (password_entry.text);
-        password_entry.text = "";
         password_entry.sensitive = false;
     }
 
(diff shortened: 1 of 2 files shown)


cassidyjames:
Rather than clear the credentials, should we just highlight them? If the user types, they'll be replaced, but otherwise if they know how they mistyped (especially if we add #352) they might have an easier time correcting it.

Philip-Scott:
Good call on the selection instead of clear, I like that experience even more :) On the manual card though, I haven't been able to trigger `wrong_username ()` after a wrong username on my local install, but I guess that we should also select either the username or password if either one is wrong?

Issue #65, opened 2017-08-13
Greeter does not respect displays settings


Corsseir:
I can not set primary monitor before sign in. I have tried copy ~/.config/monitors.xml file to /var/lib/lightdm/.config but it did not help.

I have Dell U2414H attachted by HDMI to Dell Inspirion 5378 laptop.

Corsseir:
I have changed pantheon greeter to unity greeter and bug still persist. Seems like /var/lib/lightdm/.config/monitors.xml are just ignored.

DrewBe121212:
I also experienced this issue. It seems greeter simply takes whatever bios deems as the primary monitor and uses that monitor for the login screen. I was able to rectify by physically switching which monitor was plugged into port 0 of the video card, however using what is set as the primary monitor via display settings makes far more sense.

gamunu:
I'm no expert in this subject. but Gtk 3 documentation recommends `Gdk.Display.get_primary_monitor()` over `Gdk.Screen.get_primary_monitor`. This could be the case.

https://github.com/elementary/greeter/blob/552985ec213966a89ee58e3956f8cbe8f28fbd39/src/PantheonGreeter.vala#L292

https://github.com/elementary/greeter/blob/552985ec213966a89ee58e3956f8cbe8f28fbd39/src/PantheonGreeter.vala#L281

kt1024:
1. sudo chown lightdm:lightdm /var/lib/lightdm/.config/monistors.xml
2. append "active-monitor=0" to /etc/lightdm/pantheon-greeter.conf 
both way did NOT work.

dannios:
I've got the same issue, as well as lightdm not respecting the screen _rotation_ as well.

NoisyFlake:
The workaround here works: https://gist.github.com/RyanMillerC/c48d9f63f8fe386b22a7da18cf6913d2
Though this definitely should be fixed before the final release.

lhorace:
Of course, it would be allot better that the fix mentioned by @NoisyFlake was actually incorporated in the distribution.

andreasfelix:
How should one decide which is the primary monitor if users have set different primary monitors?

Would it maybe a solution to just show the greeter on that monitor which currently has the mouse pointer?

BAProductions:
Try this instead I just tested it fixed for me.
Just make a symlink  ~/.config/monitors.xml in /var/lib/lightdm/.config/monitors.xml & done.
[...]

PR #374, opened 2019-10-28
Scale cards based on display size

### LINKED ISSUE #303: Increase user preview size on large screens
I am using a 38" ultrawide for development, and the new greeter feels out of place because it's mostly gray. It will get even lonelier without the Libvirt Guest from #296.

IMG_6416

Would it be reasonable to increase the size of the user boxes in relation to the available screen size? That way I could see more of my pretty wallpaper.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/76228809-increase-user-preview-size-on-large-screens?utm_campaign=plugin&utm_content=tracker%2F60068889&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60068889&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cassidyjames:
Fixes #303 

Right now this only scales the wallpaper on the card to be a quarter of the display size (which is roughly what the original mockup had).

Notes:
- The wallpaper on the card will match the display aspect ratio; before, it was always 2:1
- Fonts or anything else are not scaled any differently
- I'm not sure how this works on multiple differently-sized displays attached to one machine
- Right now (WIP) it's getting the display size each time to use it, need to DRY that out
- Setting the size to ¼ the display size means you can usually see 1½ users on either side, for a total of 3–5 (at least partially) visible users depending on which user you're scrolled to.


cassidyjames:
### 1024×768 (smallest supported resolution)



### 1920×1080 (same layout for 4K HiDPI)



### 2560×1440 (common on big ol' 27" displays)

cassidyjames:
It looks like the scale in the construct block is not correct, but always returns 1. The means the cards are twice as big as they should be in each dimension.

CallieG:
So, a question about setting this.   I run in 1600x900, with the "Larger" setting for text, as I have vision problems.

Would or should I be able to separately set the lightdm resolution to 1024x768, if only to make the user cards and login box large enough for me to see comfortably?

cassidyjames:
@CallieG that's not really related to this branch, as this does not affect the text size; it just makes the card itself (and wallpaper) fill up more space relative to the display.

But to answer your question, I don't believe resolution is carried over from the user session, and there's not currently a graphical way to set a manual resolution on the greeter. However, the font scaling should probably be added as a setting that gets respected on the greeter as well.

CallieG:
And Cassidy, this is why I love you guys.    I am no programmer, not even remotely, I am just an enthusiastic user, but you did actually take the time to respond to this.

And you should know that it is greatly appreciated.

BAProductions:
Not needed take look at my fix https://github.com/elementary/greeter/issues/65#issuecomment-664671246

PR #376, opened 2019-10-29
GObject-style construction

cassidyjames:
Just some cleanup for consistency.


tintou:
That's not how GObject Construction style works, subclasses are calling the construct{} block but not the `Foo ()` one. That's a no for me

Issue #120, opened 2018-06-08
support for mutter 3.29+ missing


decathorpe:
The added compositor code was probably based on gala before support for mutter 3.29+ was added, and hence support for the current development version / next stable version of mutter is missing.

meson configure output:

```
The Meson build system
Version: 0.46.1
Source dir: /builddir/build/BUILD/pantheon-greeter-3.2.0+git180608.081009.c193bacf
Build dir: /builddir/build/BUILD/pantheon-greeter-3.2.0+git180608.081009.c193bacf/x86_64-redhat-linux-gnu
Build type: native build
Project name: io.elementary.greeter
Native C compiler: cc (gcc 8.1.1 "cc (GCC) 8.1.1 20180502 (Red Hat 8.1.1-1)")
Appending CFLAGS from environment: '-O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection'
Appending LDFLAGS from environment: '-Wl,-z,relro  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld'
Native Vala compiler: valac (valac 0.40.6)
Appending LDFLAGS from environment: '-Wl,-z,relro  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld'
Build machine cpu family: x86_64
Build machine cpu: x86_64
Found pkg-config: /usr/bin/pkg-config (1.4.2)
Native dependency glib-2.0 found: YES 2.56.1
Configuring config.vala using configuration
Native dependency clutter-gtk-1.0 found: YES 1.8.4
Native dependency gdk-pixbuf-2.0 found: YES 2.36.12
Native dependency gdk-x11-3.0 found: YES 3.22.30
Native dependency gl found: YES 18.1.1
Native dependency granite found: YES 0.5
Native dependency liblightdm-gobject-1 found: YES 1.26.0
Native dependency wingpanel-2.0 found: YES 2.0
Native dependency x11 found: YES 1.6.5
Library posix found: YES
Library m found: YES
[...]

PR #377, opened 2019-10-30
Adapt to the latest Mutter API

worldofpeace:
I've tested that it builds with mutter 3.34.1 and 3.28.4.
What I've done in `SystemBackground.vala` is probably totally incorrect. A suggestion is welcome on what we actually should be doing.

What I've done is adapted the changes in https://github.com/elementary/gala/pull/566, as it's literally the code copy pasted mostly, to be used here.

Fixes https://github.com/elementary/greeter/issues/120.

It would be fabulous :v: if we could have everything adapted to latest mutter before the end of the year.

### DIFF
--- a/compositor/DBusAccelerator.vala
+++ b/compositor/DBusAccelerator.vala
@@ -44,6 +44,7 @@ namespace GreeterCompositor {
 
     public struct Accelerator {
         public string name;
+        public Meta.KeyBindingFlags grab_flags;
         public ActionMode flags;
     }
 
@@ -74,7 +75,11 @@ namespace GreeterCompositor {
             wm = _wm;
             grabbed_accelerators = new GLib.List<GrabbedAccelerator> ();
 
+#if HAS_MUTTER330
+            wm.get_display ().accelerator_activated.connect (on_accelerator_activated);
+#else
             wm.get_screen ().get_display ().accelerator_activated.connect (on_accelerator_activated);
+#endif
         }
 
         private void on_accelerator_activated (uint action, uint device_id, uint timestamp) {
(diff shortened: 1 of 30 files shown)


Tireg:
Ok now I can confirm this works great since May 8, when I dropped the first patch (https://github.com/elementary/greeter/issues/120#issuecomment-490663451).
Didn't encountered any issue from this time with greeter.
@worldofpeace has done a great work by adding this PR and adapting changes for newer mutter versions.

This would be great to have some review here as everything seems to work as expected.

worldofpeace:
> Ok now I can confirm this works great since May 8, when I dropped the first patch ([#120 (comment)](https://github.com/elementary/greeter/issues/120#issuecomment-490663451)).
> Didn't encountered any issue from this time with greeter.
> @worldofpeace has done a great work by adding this PR and adapting changes for newer mutter versions.
> 
> This would be great to have some review here as everything seems to work as expected.

Thanks @Tireg, also thanks for tracking down all the issues so carefully in gala.
It's pretty important to me as a downstream to be able to support latest mutter API's and not hold compatibility packages.

worldofpeace:
@tintou I did notice that the background with mutter 3.34 never got drawn...

worldofpeace:
Will quote that
> What I've done in SystemBackground.vala is probably totally incorrect. A suggestion is welcome on what we actually should be doing.

I believe there's something going on there.

tintou:
@worldofpeace I just pushed a fix to this, It was because `system_background.set_wallpaper ();` wasn't called in the 3.34 version, I fixed it by setting the system wallpaper at object construction as it is done in Gala

worldofpeace:
> @worldofpeace I just pushed a fix to this, It was because `system_background.set_wallpaper ();` wasn't called in the 3.34 version, I fixed it by setting the system wallpaper at object construction as it is done in Gala

Thanks :sparkles:

tintou:
Thank you :)

worldofpeace:
I'm not sure if https://github.com/elementary/gala/issues/518 will also be present in the greeter. But if any fix makes its way there it probably should be ported here too.

PR #378, opened 2019-10-31
Release 5.0.1

cassidyjames:
https://github.com/elementary/greeter/compare/5.0...master


Appeltabak:
Considering the features that are added in this release and the fact that elementary folllows semver, wouldn't this be a minor release instead of a patch release? Just wondering,

danrabbit:
I guess adding the sound settings daemon could be considered a new feature, but I think the rest of these things would qualify as just a bug fix

PR #417, opened 2020-03-13
deb-packaging: remove workaround for Ubuntu patched GSD

davidmhewitt:
Seems Ubuntu don't patch GSD as heavily anymore, so we can remove these workarounds. Will be able to remove the conditional compiles from the code in future I guess, but for now this keeps support for `bionic` too.


tintou:
Stable bionic greeter has been switched to deb-packaging-bionic.
Daily bionic greeter has been disabled. It's time to be Focal first.

PR #421, opened 2020-03-31
DBusAccelerator: Fix grab_accelerator argument with Mutter >= 3.22

tintou:
Tiny regression from https://github.com/elementary/greeter/commit/39cba2b462897dbbc10a994e9a8c7c39c4a95178

### DIFF
--- a/compositor/DBusAccelerator.vala
+++ b/compositor/DBusAccelerator.vala
@@ -44,10 +44,10 @@ namespace GreeterCompositor {
 
     public struct Accelerator {
         public string name;
+        public ActionMode flags;
 #if HAS_MUTTER332
         public Meta.KeyBindingFlags grab_flags;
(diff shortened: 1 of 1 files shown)


davidmhewitt:
@tintou I'm pretty sure this is already correct. 

Look at the `gnome-shell` code for this:
https://github.com/GNOME/gnome-shell/blob/c05098cd12d20c19e11ab66205be1db5ca426b74/js/ui/shellDBus.js#L170

Ok, their argument is named `grabFlags`, but it is the 3rd argument from the DBus call which is what we have already.

tintou:
@davidmhewitt you were right it's the third argument, our argument ordering was wrong, it is now fixed in this commit.

davidmhewitt:
@tintou We will need the same fix in gala

PR #448, opened 2020-06-30
RDNN package name

danrabbit:


davidmhewitt:
This seems right, I think. But, there are a couple of things to consider...

1) I'm not sure how the release action will handle this, it technically needs to release two packages? Equally, I'm not sure if anything needs to be done on the launchpad side.
2) Not a big deal, but we need to update seeds/metapackages so we're not including the dummy packages in the iso.
3) I appreciate this doesn't affect any GSettings etc, but when we do start to migrate GSettings in other packages to RDNN, it would be good to write some code that moves all the values to the new path. I remember back in the Loki days when we transitioned some packages, settings were lost.

jeremypw:
@danrabbit Converting to draft as there seem to be some unresolved questions relating to this.

danrabbit:
I'm fairly certain there's nothing to do here for the release action or on Launchpad. There are other repos that release multiple packages like Calendar or Music.

If this is merged, we'll want to update seeds yes but this can be done at any time

I don't think Greeter uses gsettings?

PR #455, opened 2020-08-03
compositor: Add KeyboardManager class

davidmhewitt:
I'm currently working on getting keyboard layout support working in the greeter, including getting the indicator to work and syncing the layouts from the user sessions to the lightdm session. We'll need this class from gala to listen to the GSettings key and set the right layouts.

This will have no effect on keyboard layouts in the greeter at this stage as there won't be any configured in GSettings for the `lightdm` user. I plan to use AccountsService to synchronise these in a future commit.

### DIFF
--- a/compositor/KeyboardManager.vala
+++ b/compositor/KeyboardManager.vala
@@ -109,4 +109,3 @@ namespace GreeterCompositor {
         }
     }
 }
-
--- a/compositor/KeyboardManager.vala
+++ b/compositor/KeyboardManager.vala
@@ -1,0 +1,106 @@
+//
+//  Copyright (C) 2016 Santiago León
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


compositor/KeyboardManager.vala
```diff
@@ -0,0 +1,111 @@
...
+            Object ();
+        }
+
+        construct {
+            var schema = GLib.SettingsSchemaSource.get_default ().lookup ("org.gnome.desktop.input-sources", true);
+            if (schema == null)
+                return;
+
+            settings = new GLib.Settings.full (schema, null, null);
+            Signal.connect (settings, "changed", (Callback) set_keyboard_layout, this);
```
tintou, line 48: The file really look like C code in many aspects.
Any reason to not use `settings.changed.connect (set_keyboard_layout);` ?

davidmhewitt: None other than the fact I've copied this file from Gala without changing it ;)

I'd prefer to update the code style in a separate commit.

PR #475, opened 2020-10-02
Fix compiler warnings

davidmhewitt:
`Display.get_pointer` is deprecated, so we replace this with getting the pointer device from the seat. There's a fallback to use the primary monitor here if we can't detect which monitor the mouse is on for some reason. But, I've tested this with multiple monitors and the login dialog still moves onto the monitor the pointer is on.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -307,13 +307,13 @@ public class Greeter.MainWindow : Gtk.ApplicationWindow {
     }
 
     private void maximize_window () {
-        int x, y;
         var display = Gdk.Display.get_default ();
         unowned Gdk.Seat seat = display.get_default_seat ();
         unowned Gdk.Device? pointer = seat.get_pointer ();
 
         Gdk.Monitor? monitor;
         if (pointer != null) {
+            int x, y;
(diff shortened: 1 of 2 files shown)


meisenzahl (approved):
Looks good. Also works with my multi-monitor setup.

src/MainWindow.vala
```diff
@@ -309,8 +309,17 @@ public class Greeter.MainWindow : Gtk.ApplicationWindow {
...
         int x, y;
         var display = Gdk.Display.get_default ();
-        display.get_pointer (null, out x, out y, null);
-        var monitor = display.get_monitor_at_point (x, y);
+        unowned Gdk.Seat seat = display.get_default_seat ();
+        unowned Gdk.Device? pointer = seat.get_pointer ();
+
+        Gdk.Monitor? monitor;
+        if (pointer != null) {
+            pointer.get_position (null, out x, out y);
```
tintou, line 317:
```suggestion
            int x, y;
            pointer.get_position (null, out x, out y);
```
It's probably better to have it under this scope

tintou (approved):

PR #481, opened 2020-12-03
Set XDG_CURRENT_DESKTOP to Pantheon

davidmhewitt:
To be used in conjunction with https://github.com/elementary/default-settings/pull/186/ so that we adopt the GSettings overrides, even in the case where they're specific to Pantheon.


danrabbit (approved):
Let's try it!

Issue #497, opened 2020-12-29
Open terminal from login screen


etkaar:
See also Q&A #496.

## Problem

There are cases where you would need a terminal _before_ logged in. Usually you would open a fullscreen TTY using `Ctrl+Alt+F1-F6`; however, this is not as simple as it sounds, because unfortunately in many cases this will not work due to driver problems (it will end in a simple [blackscreen](https://askubuntu.com/questions/162535/why-does-switching-to-the-tty-give-me-a-blank-screen)). There may be ways to fix that by editing or updating grub, but I personally had no chance. In that case, you need to do a more or less complex and time intense rescue to fix the problem behind.

If there is anything wrong with the login/lock screen, e.g. due to an accidentially removal of required packages, you will end up locked out of the entire system:

![locked-out](https://user-images.githubusercontent.com/40885610/103316999-81b2e980-4a2a-11eb-8fe0-69745382b04d.jpg)

The possibility to initiate a terminal from the login/lock screen would solve this and would make eOS more robust.

## Proposal

Icon and/or shortcut to open a terminal from the login/lock screen.

cassidyjames:
This seems like trying to fix a problem with a potentially bigger problem, and an invitation to way too much trouble. For extreme recovery needs (like a non-launching session), there is already recovery mode when booting and the TTY. If those don't work for some specific reason, then those should be fixed, but I don't think the solution is to add a Terminal here. Especially since the greeter is a privileged user.

PR #501, opened 2021-01-17
Add button to open terminal

meisenzahl:
Closes https://github.com/elementary/greeter/issues/497


meisenzahl:
Adds a button that should open a terminal when clicked. Unfortunately, the terminal does not open at the moment.

Bildschirmfoto von 2021-01-17 08 02 27

@danrabbit I copied your code to open `io.elementary.initial-setup` and adapted it so that `io.elementary.terminal` should be started. Unfortunately this does not work. Do you remember if you had problems back then and had to configure something special for `io.elementary.initial-setup`?

hanaral:
This is supposed to be an OSusable by anyone, adding a big ass terminal button to the bottom of the login screen is asking for trouble. How about making it a specific keybind instead?

meisenzahl:
Currently it is in the style of the other buttons like "Log in as Guest" and "Manual Login...". I am not sure what the final design will look like. But we already have keybindings to open a terminal `Ctrl+Alt+F1-F6`. So I think a button, in whatever form, makes sense.

But I'm not the right one to decide that, so I'm looking forward to feedback from @elementary/ux.

cassidyjames:
I'm really :-1: on adding a way to open Terminal from the greeter; that seems like an invitation to way too much trouble. For extreme recovery needs (like a non-launching session), there is already recovery mode when booting and the TTY.

PR #505, opened 2021-02-06
ManualCard: Modern code style and cleanup

cassidyjames:
Just some code I noticed could be dusted off when I was spelunking.

- Adds copyright header
- Pulls magic error shaking number out into a const
- Uses modern codestyle throughout
- Adds whitespace for legibility
- Uses `Granite.STYLE_CLASS_ROUNDED` and bumps Granite dep to match
- Uses weak reference to `Gtk.StyleContext` 
- Re-orders and groups the style contexts to keep things organized

### DIFF
--- a/src/Cards/ManualCard.vala
+++ b/src/Cards/ManualCard.vala
@@ -1,3 +1,22 @@
+/*
+ * Copyright 2018–2021 elementary, Inc. (https://elementary.io)
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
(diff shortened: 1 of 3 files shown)


src/Cards/ManualCard.vala
```diff
@@ -1,62 +1,90 @@
...
-        form_grid.orientation = Gtk.Orientation.VERTICAL;
-        form_grid.column_spacing = 6;
-        form_grid.row_spacing = 12;
-        form_grid.margin = 24;
-        form_grid.attach (icon, 0, 0, 2, 1);
-        form_grid.attach (label, 0, 1, 2, 1);
+        var form_grid = new Gtk.Grid () {
+            column_spacing = 6,
+            margin = 24,
+            orientation = Gtk.Orientation.VERTICAL,
```
danrabbit, line 71: I'm actually not sure why we're setting orientation here if we're using `attach`. It kind of doesn't matter if we're not using `add`

danrabbit (approved):
Lovin' the cleanup branches. Made a tiny comment, but it's not blocking. Nice work!

src/Cards/ManualCard.vala
cassidyjames: @danrabbit yeah I bet it predated the session button or something and just wasn't removed. Addressed!

PR #506, opened 2021-02-06
UserCard: Code cleanup

cassidyjames:
Complimentary to #505. Looks like more than it is because it wraps some long lines for legibility.

### DIFF
--- a/src/Cards/BaseCard.vala
+++ b/src/Cards/BaseCard.vala
@@ -28,6 +28,8 @@ public abstract class Greeter.BaseCard : Gtk.Revealer {
     public bool need_password { get; set; default = false; }
     public bool use_fingerprint { get; set; default = false; }
 
+    public const int ERROR_SHAKE_DURATION = 450;
+
     static construct {
         css_provider = new Gtk.CssProvider ();
         css_provider.load_from_resource ("/io/elementary/greeter/Card.css");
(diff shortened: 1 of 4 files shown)


cassidyjames:
Not sure why I didn't mark this as ready earlier. Just code style cleanup, no functional change.

src/Cards/BaseCard.vala
```diff
@@ -28,6 +28,8 @@ public abstract class Greeter.BaseCard : Gtk.Revealer {
     public bool need_password { get; set; default = false; }
     public bool use_fingerprint { get; set; default = false; }
 
+    public const int ERROR_SHAKE_DURATION = 450;
```
danrabbit, line 31: Did you intend to use this somewhere else? Why did this move from ManualCard? And does it need to be `public` or was this intended to be `protected`?

cassidyjames: This const is used in both UserCard and ManualCard; in the ManualCard cleanup I made it a const, but since UserCard also needs it, I moved it to BaseCard. But yeah shouldn't need to be public.

```suggestion
    protected const int ERROR_SHAKE_DURATION = 450;
```

danrabbit (approved):
Looks good :+1:

danrabbit:
I lied, it doesn't look good

danrabbit:
Okay so, it did look good, but CI didn't agree. But now it does. :point_right: :sunglasses: :point_right:

PR #509, opened 2021-02-20
Handle installer/demo session mode switching

davidmhewitt:
To be used in conjunction with https://github.com/elementary/installer/pull/486

To facilitate allowing both the installer session and live (demo) session to use the same user, so that we don't have issues with permissions by using a guest user for the demo session, we can have the greeter do the switching between the two session types.

Unfortunately the `SwitchToUser` method on the lightdm DBus interface doesn't work for switching between two sessions for the same user, so it seems the best way that the installer can signal to the greeter that it wants the demo session is to touch a file and then log out, allowing the greeter to automatically log back in to the correct session.

If this file is not present, the greeter will just launch the installer session.


src/MainWindow.vala
```diff
@@ -405,7 +407,21 @@ public class Greeter.MainWindow : Gtk.ApplicationWindow {
 
             var action_group = get_action_group ("session");
             try {
-                lightdm_greeter.start_session_sync (action_group.get_action_state ("select").get_string ());
+                var session = action_group.get_action_state ("select").get_string ();
```
tintou, line 410:
```suggestion
                unowned var session = action_group.get_action_state ("select").get_string ();
```

tintou (approved):

PR #537, opened 2021-06-17
MainWindow: Use scroll and swipe handling from Hdy

danrabbit:
Replaces custom scroll handling with scroll and swipe handling native to Hdy.Carousel. This both gives us 1:1 swiping and long swiping across several cards

The previous method here relied on packing hidden cards into the carousel and hiding and showing all cards for the manual card entry. This doesn't work because you can swipe over to hidden cards, which is bad. So instead, use a Gtk.Stack to switch between manual login and user cards

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -69,6 +69,7 @@ public class Greeter.MainWindow : Gtk.ApplicationWindow {
         extra_login_grid.valign = Gtk.Align.END;
         extra_login_grid.column_spacing = 12;
         extra_login_grid.column_homogeneous = true;
+        extra_login_grid.attach (manual_login_button, 1, 0);
 
         try {
             var gtksettings = Gtk.Settings.get_default ();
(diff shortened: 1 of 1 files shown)


src/MainWindow.vala
```diff
@@ -88,22 +88,26 @@ public class Greeter.MainWindow : Gtk.ApplicationWindow {
         var manual_card = new Greeter.ManualCard ();
 
         carousel = new Hdy.Carousel () {
-            interactive = false,
+            allow_long_swipes = true,
```
cassidyjames, line 91: I'm mixed on this; it feels really imprecise and finicky to me, but I can see how swiping repeatedly through many users would also be tiring/annoying.

cassidyjames (approved):
Nice straightfoward improvement. My only questions are related to the long swipes (see previous comment) and switching between cards with the keyboard: this branch seems to slow that down, but I think we need to improve how that's being handled anyway. So I'm :+1: to merge this and continue iterating.

vjr:
Does this PR preserve the mouse wheel based scrolling ability?

JoseExposito (approved):
As smooth as silk! Nice job :tada: 

I also had doubts about long swipes, but after testing with 3 accounts it really is an improvement, but left that decision to Cassidy. I don't know how common is having more that 1 or 2 accounts, though.

@vjr yes, both mouse wheel and keyboard navigation keep working.

PR #538, opened 2021-06-18
Set users' preferred accent color

### LINKED ISSUE #470: Greeter cards should use the selected user’s accent
Just thought of this, I’m pretty sure it’s possible

danrabbit:
Fixes #470

### DIFF
--- a/src/Cards/UserCard.vala
+++ b/src/Cards/UserCard.vala
@@ -29,6 +29,7 @@ public class Greeter.UserCard : Greeter.BaseCard {
     public double reveal_ratio { get; private set; default = 0.0; }
     public bool is_24h { get; set; default = true; }
 
+    public int prefers_accent_color { get; set; default = 6; }
     public int sleep_inactive_ac_timeout { get; set; default = 1200; }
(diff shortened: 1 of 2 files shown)


cassidyjames (approved):
This is pretty straightforward! The only case it doesn't account for is automatic based on wallpaper, which currently will default to `blueberry`. That's fine at least to start, as I'm not sure how we'd solve that.

src/Cards/UserCard.vala
```diff
@@ -280,11 +284,50 @@ public class Greeter.UserCard : Greeter.BaseCard {
...
+        });
+
         grab_focus.connect (() => {
             password_entry.grab_focus_without_selecting ();
         });
     }
 
+    private void update_style () {
+        var gtksettings = Gtk.Settings.get_default ();
+        gtksettings.gtk_icon_theme_name = "elementary";
```
cassidyjames, line 299: I guess we probably don't need to be doing this every time?

danrabbit: Yeah I actually don't know why that was there. Seems to work fine without it

PR #542, opened 2021-07-05
meson: require libhandy-1 >= 1.1.90

decathorpe:
HdyCarousel allow-long-swipes property was added in version 1.1.90:
https://gitlab.gnome.org/GNOME/libhandy/-/blob/master/NEWS#L130


Conan-Kudo (approved):

danrabbit (approved):
Thanks for noticing that!

PR #566, opened 2021-09-17
UserCard: use accent color for logged_in check

### LINKED ISSUE #555: Tick Color for Active User Doesn't Use Active User's Preferred Color
## What Happened
The password box accent color seems to adopt the preferred color of the current active user, while the tick (identifier of current active user) does not. See this picture:
image

## Expected Behavior
Same color for both elements.

## Steps to Reproduce
1. Log in as any user (not guest).
2. Lock computer (Superkey + L).
3. Observe Greeter screen.

## Platform Information
Elementary OS 6

danrabbit:
Fixes #555

### DIFF
--- a/data/Check.css
+++ b/data/Check.css
@@ -1,0 +1,11 @@
+check {
+    background-color: @accent_color;
+    border-radius: 99px;
+    color: white;
+    margin: 2px;
+    min-height: 20px;
+    min-width: 20px;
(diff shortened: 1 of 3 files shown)


cassidyjames (approved):
Looks nice. I imagine we may want to move these styles to the stylesheet if we're gonna solve this same issue in Files and the Appearance plug, but this is a start.

PR #587, opened 2021-12-07
Update README.md

marmeladapk:
Updated required dependencies in readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,6 @@ You'll need the following dependencies:
 * libhandy-1-dev >= 0.90.0
 * liblightdm-gobject-1-dev
 * libmutter-6-dev
-* libwingpanel-3.0-dev
 * libx11-dev
 * meson
 * valac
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,8 +15,8 @@ You'll need the following dependencies:
 * libgtk-3-dev
 * libhandy-1-dev >= 0.90.0
 * liblightdm-gobject-1-dev
-* libmutter
-* libwingpanel-2.0-dev
+* libmutter-6-dev
+* libwingpanel-3.0-dev
```
danrabbit, line 19: I think we actually dropped the libwingpanel dep completely since we use the panel as its own process now

danrabbit (changes requested):
Hey thanks for noticing this! Just one comment

README.md
marmeladapk: I checked and removed wingpanel-dev package, greeter still built.

danrabbit (approved):
thanks!

Issue #578, opened 2021-11-04
Greeter does not load correctly on boot


damiante:
### What Happened?

When booting, some issue occurs between lightdm and greeter. I often see Greeter load with the lock screen for a brief moment, but it will crash to a black screen with a blinking cursor or my motherboard's manufacturer logo. From this screen TTY can be enabled with CTRL + ALT + F2 and restarting lightdm _can_ solve the issue but does not consistently do so. This occurs on almost all cold boots and most reboots. After successfully launching the service everything works correctly.

Please refer to this stack overflow topic: https://elementaryos.stackexchange.com/questions/29090/how-to-make-lightdm-more-stable-on-boot
There are some logs provided in the topic

Initial suspicion is that this is config-related but I don't have much to go on. Happy to provide additional logs if asked.

### Steps to Reproduce

1. Boot system
2. Usually, I will land on a screen with my motherboard's logo
3. Enter TTY with CTRL + ALT +F2 and check `/var/log/syslog` for evidence that lightdm has failed on startup
4. do `sudo service lightdm restart`; this may require several attempts
5. Greeter initialises and the OS/GUI can be used as normal

### Expected Behavior

The system correctly launches lightdm + greeter on every boot

### OS Version

6.x (Odin)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
Contents of doing `sudo lightdm --test-mode --debug`: 
https://wtools.io/paste-code/b75Y

Additional logs provided on request
```

### Hardware Info

output of `lshw -short`

```
H/W path                  Device          Class          Description
====================================================================
                                          system         System Product Name (SKU)
/0                                        bus            TUF GAMING B550M-PLUS
/0/0                                      memory         64KiB BIOS
/0/2c                                     memory         16GiB System Memory
/0/2c/0                                   memory         Project-Id-Version: lshwReport-Msgid-Bugs-To: FULL NAME 
/0/2c/1                                   memory         8GiB DIMM DDR4 Synchronous Unbuffered (Unregistered) 2400 MHz (0.4 ns)
/0/2c/2                                   memory         Project-Id-Version: lshwReport-Msgid-Bugs-To: FULL NAME 
/0/2c/3                                   memory         8GiB DIMM DDR4 Synchronous Unbuffered (Unregistered) 2400 MHz (0.4 ns)
/0/2f                                     memory         384KiB L1 cache
/0/30                                     memory         3MiB L2 cache
/0/31                                     memory         32MiB L3 cache
[...]

PR #598, opened 2022-02-24
Modifying the logged_in variable to be class private to allow access …

oddveryodd:
…outside of the function it is allocated in.  The logged_in variable is referenced by a different function, and a race condition can occur, causing a de reference to a freed element.


meisenzahl:
Closes #578

meisenzahl (approved):
I can confirm that this fixes a random issue on my work laptop that loads Greeter but prevents the login cards from showing.

meisenzahl:
@oddveryodd thank you for your contribution. Great work!

PR #612, opened 2022-05-23
Mainwindow: launch installer on load

### LINKED ISSUE #600: Latest ISO doesn't automatically launch installer
### What Happened?

The greeter is supposed to be responsible for launching the installer session when it detects it is installed.

Something has changed in the stack somewhere that prevents this from occurring successfully.

I believe this is because the `elementary` user that `casper` creates during bootup of the live ISO isn't recognised as a user by LightDM, and the Greeter follows the "no users, so launch initial setup" path.

It seems that `casper` creates this user as UID 999, and probably always has. LightDM uses AccountsService to get the list of valid "human" users to be displayed in the greeter. It seems that AccountsService now only treats users with UID >= 1000 as real users, whereas previously it must have done something else. So, this `elementary` user is not presented to the greeter as a valid user, and the initial setup flow is launched instead of the installer.

I don't think we can just change the logic to just launch the installer session regardless of whether there's a user or not, because it relies on launching the installer session _as_ that user.

So, possible solutions:
* Patch `casper` to create the live user with UID `1000`
* Patch whichever config AccountsService is looking at to have valid UIDs as >= 999 (probably in the ISO only)
* Completely rethink/rework how the installer "sessions" work (i.e. have the installer run in a pantheon session, or see if it's possible to run it directly in the greeter instead of its own session)

### Steps to Reproduce

1. Boot 20220228 ISO
2. End up at initial setup instead of installer

### Expected Behavior

1. Boot ISO
2. Get to installer

### OS Version

7.x (Early Access)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

danirabbit:
Fixes #600

Tested this in a live session and it seems to work :shrug:

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -446,7 +446,7 @@ public class Greeter.MainWindow : Gtk.ApplicationWindow {
             try {
                 lightdm_greeter.authenticate ("elementary");
             } catch (Error e) {
-                critical ("Couldn't launch installer session: %s", e.message)
+                critical ("Couldn't launch installer session: %s", e.message);
             }
         }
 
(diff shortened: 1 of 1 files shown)


davidmhewitt:
We worked around this in a different way by patching `casper`, so this can be closed.

Issue #610, opened 2022-05-23
Screen reader shortcut doesn't launch


danirabbit:
### What Happened?

The screen reader shortcut Super + Alt + S doesn't enable the screen reader in either the Greeter or Installer (which uses the greeter compositor, so I think it's related?)

### Steps to Reproduce

1. Boot from install media
2. Attempt to enable screen reader with Super + Alt + S
3. Open the a11y indicator to see that the screen reader is not enabled

### Expected Behavior

When pressing the keyboard shorcut Super + Alt + S, the screen reader should be toggled just as it is during a normal session

### OS Version

7.x (Early Access)

### Software Version

Compiled from git

lenemter:
I looked into this a while ago, and seems like the issue is in gnome-settings-daemon, its "screenreader" shortcut doesn't work in greeter session. I think we can handle this shortcut by ourselves in compositor or in greeter itself.

PR #614, opened 2022-06-02
DBusAccelerator: Use the right prototype to expose the interface

tintou:


danrabbit:
Is this meant to fix #610?

tintou:
This has been part of my investigation yes but it doesn't fix this issue unfortunately

danrabbit (approved):

PR #549, opened 2021-08-15
Blurred Pixbuf on Clutter Canvas

torikulhabib:
Another way to paint picture blurred on stage

PR #623, opened 2022-10-31
Blurred background clutter canvas

torikulhabib:
replace https://github.com/elementary/greeter/pull/549


vjr:
Hi @torikulhabib  is this the blurred greeter wallpaper background?

I like the idea +1 from me but I might only be able to try it this weekend

compositor/SystemBackground.vala
```diff
@@ -33,26 +37,106 @@ public class Greeter.SystemBackground : GLib.Object {
...
+    public async void set_wallpaper (string path) throws Error {
+        if (path == null) {
+            path = DEFAULT_GRAY_BACKGROUND;
+        } else if (path == "") {
+            path = DEFAULT_BACKGROUND_PATH;
+        }
+
+        if (path != DEFAULT_GRAY_BACKGROUND && GLib.FileUtils.test (path, GLib.FileTest.EXISTS)) {
+            var pixbuf = new Gdk.Pixbuf.from_file (path);
+            var surface = new Granite.Drawing.BufferSurface (pixbuf.width, pixbuf.height);
```
danirabbit, line 53: `Granite.Drawing.BufferSurface` was removed from Granite 7, so if you want to use it here you should add it as we did in Gala: https://github.com/elementary/gala/pull/1428

danirabbit (changes requested):
Left a couple of comments

Another problem I'm noticing in this branch is that the panel seems to be opaque when it would otherwise be transparent without blur. This seems counter-intuitive

I would also suggest increasing the blur radius. To my eye, `10` looks more like poor upscaling than an intentional blur

compositor/SystemBackground.vala
torikulhabib: ok

vjr:
Hi @torikulhabib I tried this PR in VM and also my actual v7 installation and I like the concept but some observations for improvements...

* The background is not scaled correctly on my HiDPI monitor - it's too large - only part of the background is visible.
* The fade should be a "crossfade" or "blend" I guess from one user's background to the other user's background - currently upon switching the user it starts from blank and fades-in.
* The performance is jerky (not smooth) and this is probably due to the code trying to iterate the `fadebg` steps from 1 to 10 "manually" instead of I guess animating it?

I'm not familiar with the code to do pixbuf crossfading - could you look at the code in gala and try something similar here? If you see the wallpaper fading when you change your wallpaper in the desktop settings, it's very smooth and pleasant - if you try to re-use that code the greeter would look nice too.

Thanks for this PR! Personally I hope this idea gets implemented in the greeter :-)

vjr:
@torikulhabib the scaling is fixed but I would still suggest trying to steal the code from gala and modify it a little to add the crossfade and blur? I might try it myself but not too familiar with UI coding.

compositor/SystemBackground.vala
```diff
@@ -33,26 +37,122 @@ public class Greeter.SystemBackground : GLib.Object {
...
+                    remove_time = fadebg = 0;
+                    remove_time = Timeout.add (50, ()=> {
+                        re_draw ();
+                        if (fadebg > 10) {
+                            remove_time = 0;
+                            return false;
+                        }
+                        fadebg++;
+                        return true;
+                    });
```
tintou, line 71: Instead of fading manually, you have to use the `Clutter.Transition` API bound the the Actor

tintou (changes requested):

BAProductions:
No working For Me

BAProductions:
> replace #549
I just made a pr agent your repos to fix the wallpaper no showing up on my VM.

PR #639, opened 2023-03-04
build: require mutter 42

tintou:
Same as https://github.com/elementary/gala/pull/1496


bobby285271 (approved):
Builds and works for me, mutter 42 & 43.

> Same as elementary/gala#1496

My PR is missing some [file rename](https://github.com/elementary/greeter/compare/053932d44a7897add0d22277a2280a0ce20c310b..326515e3d39b82e7e484d4f9c82367c4a78c648f) though, we can sync back if this is preferred

tintou:
Yeah I prefer linking to a single file so that it is easier to diff the file :)

PR #643, opened 2023-03-26
Add support for Mutter 44

bobby285271:
Some copy paste of https://github.com/elementary/gala/pull/1570.


pimvullers:
Tested on Gentoo Linux, build passing, seems to function as expected. (Included as patch in my Gentoo ebuild.)

PR #649, opened 2023-05-10
Fix meson warnings

lenemter:


meson.build
```diff
@@ -2,7 +2,7 @@ project(
     'io.elementary.greeter',
     'vala', 'c',
     version: '6.1.1',
-    meson_version: '>= 0.49.0'
+    meson_version: '>= 0.58.0'
```
ryonakano, line 5: I guess the minimum version of meson would be 0.56.0.

---

`source_root()` was deprecated and `project_source_root()` was introduced in  0.56.0:

https://mesonbuild.com/Reference-manual_builtin_meson.html#mesonsource_root
https://mesonbuild.com/Reference-manual_builtin_meson.html#mesonproject_source_root

`get_pkgconfig_variable()` was deprecated in  0.56.0:

https://mesonbuild.com/Reference-manual_returned_dep.html#depget_pkgconfig_variable

`get_variable()` was introduced in 0.51.0:

https://mesonbuild.com/Reference-manual_returned_dep.html#depget_variable

ryonakano: Sorry, looks like it's fine with 0.58.0. I got the following error when I change this line to 0.56.0:

```
WARNING: Project specifies a minimum meson_version '>= 0.56.0' but uses features which were added in newer versions:
 * 0.58.0: {'Positional argument to dep.get_variable()'}
```

ryonakano (approved):

PR #658, opened 2023-06-15
Use user's xkb options

### LINKED ISSUE #541: Some shortcuts to switch keyboard layout does not work on lockscreen
## What Happened
With some shortcuts (detailed below), when pressing the shortcut to switch keyboard layout (as set in Settings > Keyboard > Layout), nothing happens and keyboard layout does not switch. Instead I have to click on the Wingpanel keyboard layout indicator and switch the layout slowly like that.

## Expected Behavior
The switch keyboard layout shortcut should work in the lockscreen like it works in the rest of the system after logging in.

## Steps to Reproduce
1. Set any of the following shortcuts in `Settings > Keyboard > Layout` to switch layout: `caps lock`, `shift+ caps lock`, `alt + caps lock`, `alt + space`, `ctrl + alt`, `ctrl + shift`
2. Lock the session
3. Wake the computer and enter the greet / lock screen 
4. Press the shortcut you set before, layout does not change

## Platform Information
Odin 6.0 Beta

lenemter:
Xkb options are responsible for "switch layout" shortcut, Compose key and Caps Lock behavior.

Requires https://github.com/elementary/settings-daemon/pull/67

Fixes #541

### DIFF
--- a/src/Cards/UserCard.vala
+++ b/src/Cards/UserCard.vala
@@ -41,6 +41,7 @@ public class Greeter.UserCard : Greeter.BaseCard {
     private Gtk.Revealer form_revealer;
     private Gtk.Stack login_stack;
     private Greeter.PasswordEntry password_entry;
+    private Gtk.Label username_label;
 
     private SelectionCheck logged_in;
     private unowned Gtk.StyleContext logged_in_context;
@@ -52,7 +53,7 @@ public class Greeter.UserCard : Greeter.BaseCard {
     construct {
         need_password = true;
 
-        var username_label = new Gtk.Label (lightdm_user.display_name) {
+        username_label = new Gtk.Label (lightdm_user.display_name) {
             hexpand = true,
             margin = 24,
             margin_bottom = 12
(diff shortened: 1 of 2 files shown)


Marukesu:
i think we should use the `InputSources` property provided in the `org.freedesktop.Accounts.User` interface and remove the related keys in the pantehon and settings daemon interfaces (why we duplicate the keyboard layout info?).

zeebok:
Maybe I am misunderstanding this PR or testing in VM is causing odd behavior, but the keyboard shortcuts such as compose key and switch layout do not work. At least when set to Caps Lock, that key's behavior remains default.

lenemter:
@zeebok Make sure you installed settings-daemon from git. Also, there is some bug with git version, it no longer autostarts for me after installing, so launch it manually from terminal. Hope it will make it work for you

zeebok (approved):
Alright was able to test it correctly and confirm it works. Testing did reveal that if you have the Composer Key as Caps Lock, the alt+caps lock switch layout shortcut no longer works, but that is a bug for.... some other repo.

There is a conflict with master to fix, and the other comment from Marukesu to address, but this does work and look good.

lenemter:
@Marukesu I think this is out of scope of this PR. Can you file a new issue about duplicating keyboard info? Also, where do you find about this `InputSources` in `org.freedesktop.Accounts.User`. I tried to google anything, but found nothing about this.

PR #665, opened 2023-06-19
Release 7.0.0

danirabbit:


lenemter:
We need to update version number in the meson file as well :)

danirabbit:
@lenemter good catch, thank you!

lenemter (approved):

PR #667, opened 2023-06-26
Improve handling of users' accent color

lenemter:
* Set user's accent color globally using `org.gnome.desktop.interface gtk-theme`
* Style each user's check individually

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -28,6 +28,7 @@ public class Greeter.MainWindow : Gtk.ApplicationWindow {
     private Hdy.Carousel carousel;
     private LightDM.Greeter lightdm_greeter;
     private Greeter.Settings settings;
+    private Gtk.Button guest_login_button;
     private Gtk.ToggleButton manual_login_button;
     private Greeter.DateTimeWidget datetime_widget;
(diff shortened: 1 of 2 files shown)


src/MainWindow.vala
```diff
@@ -70,15 +71,9 @@ public class Greeter.MainWindow : Gtk.ApplicationWindow {
         extra_login_grid.column_spacing = 12;
         extra_login_grid.column_homogeneous = true;
 
-        try {
-            var gtksettings = Gtk.Settings.get_default ();
-            gtksettings.gtk_icon_theme_name = "elementary";
```
danirabbit, line 75: We added in the hardcoded icon theme here because some people had broken their panel icons etc

danirabbit (commented):
This looks good! Very cool getting the right accent color for the panel too. Nice job. Just one comment

src/MainWindow.vala
lenemter: This sets an icon theme only for the greeter app itself, since it uses Gtk.Settings, doesn't it?

danirabbit: Oh yeah hm :shrug:

danirabbit (approved):
:tada:

PR #668, opened 2023-06-29
Fix screenreader shortcut

### LINKED ISSUE #610: Screen reader shortcut doesn't launch

lenemter:
Fixes #610 
Requires https://github.com/elementary/wingpanel-indicator-a11y/pull/56

### DIFF
--- a/compositor/BackgroundUtils.vala
+++ b/compositor/BackgroundUtils.vala
@@ -22,7 +22,7 @@
  *   related to it are copied from Gala.DBus.
  */
 
- namespace BackgroundUtils {
+namespace BackgroundUtils {
     private const double SATURATION_WEIGHT = 1.5;
     private const double WEIGHT_THRESHOLD = 1.0;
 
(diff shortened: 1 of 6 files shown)


lenemter:
Drafting because this PR requires some work in a11y indicator

danirabbit (approved):
Works for me! Nice job :tada:

PR #671, opened 2023-06-30
Use settings-daemon background files

lenemter:
For https://github.com/elementary/settings-daemon/pull/72

### DIFF
--- a/src/Cards/UserCard.vala
+++ b/src/Cards/UserCard.vala
@@ -161,28 +161,25 @@ public class Greeter.UserCard : Greeter.BaseCard {
             GLib.BindingFlags.SYNC_CREATE
         );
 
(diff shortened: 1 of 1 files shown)


Marukesu:
So, there's a `BackgroundFile` property in the `org.freedesktop.Accounts.User` interface (i believe this is one of the ubuntu's patches, since i don't see it in the upstream XML), and in the `org.freedesktop.DisplayManager.AccountsService` one (this one is provided by LightDM i believe).

we should be using one of those instead of hardcoding a path.

lenemter:
@Marukesu I'll use `org.freedesktop.DisplayManager.AccountsService` because greeter already supports it, I just removed it here.

This requires some changes in https://github.com/elementary/settings-daemon/pull/72 and this PR can be closed.

PR #673, opened 2023-07-01
Add auto accent color based on wallpaper

### LINKED ISSUE #569: Accent color not applied to lock/login screens if using adaptive color
### What Happened?

When using an adaptive accent color, the default blue tick mark and password field accent is used instead of the automatically chosen accent color.

### Steps to Reproduce

1. Set accent color to adaptive
2. Lock screen, look at tick and password field accent color
3. Reset to a specific color to compare

### Expected Behavior

Would use the uniform accent color to the rest of the system, like it would if an accent were manually chosen.

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

lenemter:
Fixes #569

### DIFF
--- a/src/AccentColor/AccentColorManager.vala
+++ b/src/AccentColor/AccentColorManager.vala
@@ -1,0 +1,138 @@
+/*
+ * Copyright 2021-2023 elementary, Inc. <https://elementary.io>
+ * SPDX-License-Identifier: GPL-3.0-or-later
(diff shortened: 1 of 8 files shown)


Marukesu:
So, i think a better way to do this is to have the SettingsDaemon provide the accent color (in the same format as the portal, instead of a enum with the selected accent), then Greeter and Gala would only need to get the approximate theme. (in fact, there's elementary/gala#1674 about using the portal for the theme accent there).

PR #696, opened 2024-01-28
Cleanup

lenemter:
* Avoid syncing animations (not sure how they might get unsynced)
* Avoid using actions
* Get rid of `index_delta` and `next_delta` stuff
* Cleanup keypress handling

### DIFF
--- a/src/Cards/UserCard.vala
+++ b/src/Cards/UserCard.vala
@@ -12,7 +12,6 @@ public class Greeter.UserCard : Greeter.BaseCard {
 
     public LightDM.User lightdm_user { get; construct; }
     public bool show_input { get; set; default = false; }
-    public double reveal_ratio { get; private set; default = 0.0; }
     public bool is_24h { get; set; default = true; }
 
     public int prefers_accent_color { get; set; default = 6; }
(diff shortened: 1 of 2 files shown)


src/Cards/UserCard.vala
```diff
@@ -236,20 +235,10 @@ public class Greeter.UserCard : Greeter.BaseCard {
             }
         });
 
-        // This makes all the animations synchonous
-        form_revealer.size_allocate.connect ((alloc) => {
```
danirabbit, line 240: I think originally this was meant to sync the animation of the entry being revealed with the animation of the card switching. But it seems like maybe we broke that. Not sure if it's worth trying to fix instead?

danirabbit (changes requested):
Can you resolve conflicts please? :)

src/Cards/UserCard.vala
lenemter: I think it's a tricky thing to do with Gtk.Revealer + Hdy.Carousel. Maybe worth looking into this after Gtk4 port?

lenemter:
@danirabbit Done

danirabbit (approved):
:rocket:

PR #697, opened 2024-01-29
Compositor: Add fade-in animation

lenemter:
Closes https://github.com/elementary/greeter/issues/342
Fixes https://github.com/elementary/greeter/issues/485
Fixes https://github.com/elementary/greeter/issues/264

PR #698, opened 2024-02-03
Add blurred background

lenemter:
Closes https://github.com/elementary/greeter/pull/358
Closes https://github.com/elementary/greeter/pull/623

Background code was copied from Gala


vjr:
I tried this branch and it works well for me, so thumbs up from my side. Hopefully this is reviewed/merged :-)

vjr:
> I tried this branch and it works well for me, so thumbs up from my side. Hopefully this is reviewed/merged :-)

Scratch that! This PR seems to break/interfere with the "fade in" PR #697 recently merged. I was looking at a problem on my machine (with AMD RX580 GPU on OS 7.1 daily) with the fade-in being all noise/garbled but only the wallpaper/background, the other elements (ui_group) fade-in properly. I have not yet tested with OS8 ISO.

lenemter:
@vjr Oops, I'll fix the fade-in animation as soon as I can

danirabbit:
@vjr since we reverted the fade in for now, does this still break for you?

vjr:
@danirabbit yes the blank screen issue is still present.

I just now tested on an updated os8 install showing greeter version `7.0.0 r1550 pkg96 daily ubuntu8.1`

Same issue after building and installing this branch so I don't think it was the fade-in PR so hopefully it will return in some form because it's cool.

My bad if I misguided y'all with my recent comments about the fadein causing artefacts which was occurring only on my older PC with an AMD RX 580 GPU.

vjr:
Okay I just tested this branch with one change to move the `Gtk.init` call a few lines lower in `src/Application.vala` like was done in the fade in PR and it fixes the blank screen issue.

danirabbit (commented):
A problem this introduces is now for light wallpapers the panel becomes illegible. So we probably need to also include everything for setting the panel for light or dark and probably sync the panel transparency setting for the user as well :sweat_smile:

lenemter:
@danirabbit I fixed panel contrast here. I'll prepare a branch for syncing the transparency setting later

danirabbit (changes requested):
Looks so good, couple of small things:

- [ ] It looks like the panel transition doesn't start until after the wallpaper transition completes. Is there a way to make this happen at the same time? Basically what I'm seeing is you switch cards, the wallpaper transition happens, then the panel transition happens.
- [ ] It looks like this is being set/updated when cards are switched, but it isn't set when the greeter first loads. So if the user that is focused when the greeter starts has a light wallpaper, the icons will be white until you focus another card and then switch back

danirabbit (approved):
LFG :rocket:

PR #702, opened 2024-02-08
Revert "Revert "Compositor: Add fade-in animation""

danirabbit:
Reverts elementary/greeter#701

Okay so this actually made things worse for me somehow and @vjr commented on another PR:

> Same issue after building and installing this branch so I don't think it was the fade-in PR so hopefully it will return in some form because it's cool.

So I don't think this caused the dark screen issue thing

### DIFF
--- a/compositor/WindowManager.vala
+++ b/compositor/WindowManager.vala
@@ -76,6 +76,21 @@ namespace GreeterCompositor {
         public override void start () {
             show_stage ();
 
+            unowned var background_actor = system_background.background_actor;
+            background_actor.opacity = 0;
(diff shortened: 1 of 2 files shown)


lenemter (approved):

lenemter:
@vjr Do you have any glitches with `master` version currently?

vjr:
@lenemter no glitches with master. Note I only get the glitch with this fade in code on an AMD RX 580 GPU on my desktop PC, it works fine on my laptop though. It could just be an external issue maybe amdgpu driver or Mesa etc so it's ok to merge this pr.

PR #703, opened 2024-02-09
Simpler fade-in animation

lenemter:
Closes https://github.com/elementary/greeter/pull/702

Uses simpler way to animate fade-in. Should not produce any glitches whatsoever.

### DIFF
--- a/compositor/WindowManager.vala
+++ b/compositor/WindowManager.vala
@@ -46,7 +46,7 @@ namespace GreeterCompositor {
 
         public Greeter.SystemBackground system_background { get; private set; }
 
-        private Clutter.Actor? fade_in_screen = null;
+        private Clutter.Actor fade_in_screen;
 
         private Meta.PluginInfo info;
 
--- a/compositor/WindowManager.vala
+++ b/compositor/WindowManager.vala
@@ -46,7 +46,9 @@ namespace GreeterCompositor {
 
         public Greeter.SystemBackground system_background { get; private set; }
 
-        Meta.PluginInfo info;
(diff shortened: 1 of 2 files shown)


vjr:
I can test this branch and comment later tonight which is about 4 or 5 hours from now.

vjr:
OK I've observed the following:

This PR works with OS 8 daily on my newer laptop (Ryzen 7840U with 780M iGPU) but not on my older desktop with an AMD RX 580 GPU - the "blank screen" issue persists - actually you can see the user cards stuck in shrunken size very faint though - and just typing password I can login. I wonder if it's something to do with Noble still being in development.

This PR works well on OS 7.1 daily on my older desktop with the AMD RX 580 GPU (where the OS 8 install has the blank screen issue). I do not have OS 7.1 daily on my laptop to test but I have another Ryzen 6800U laptop with 680M iGPU which I can install both daily ISOs of OS 7.1 and 8 and test if not tonight then over the weekend.

I think this PR is OK to merge IMHO both code and behaviour LGTM but I'll leave it to others to also review/test.

compositor/WindowManager.vala
```diff
@@ -46,7 +46,9 @@ namespace GreeterCompositor {
 
         public Greeter.SystemBackground system_background { get; private set; }
 
-        Meta.PluginInfo info;
+        private Clutter.Actor? fade_in_screen = null;
```
lenemter, line 49:
```suggestion
        private Clutter.Actor fade_in_screen;
```

danirabbit (approved):
:heart:

PR #684, opened 2023-09-02
Support dark mode

lenemter:
Fixes https://github.com/elementary/greeter/issues/688
Closes #563
Closes #499

To do:
- [x] Deduplicate objects
- [x] More testing

PR #705, opened 2024-02-14
Fix criticals

lenemter:
Should fix some criticals in https://github.com/elementary/greeter/pull/684#issuecomment-1942125252


src/Cards/UserCard.vala
```diff
@@ -434,6 +434,14 @@ public class Greeter.UserCard : Greeter.BaseCard {
         settings.set_value ("xkb-options", options);
     }
 
+    private void set_or_reset_settings_key (GLib.Settings settings, string key, GLib.Variant? value) {
```
danirabbit, line 437: Could you add a code comment here about why we want to reset rather than just doing nothing? Also is there a reason we want to reset only these keys and not others?

danirabbit (approved):

PR #707, opened 2024-02-16
BackgroundManager: send `changed` signal immediately

lenemter:
This change was requested in https://github.com/elementary/greeter/pull/698#pullrequestreview-1883840770

This should be tested and integrated into Gala/Wingpanel first.


danirabbit (approved):
Hell yeah. Works the same as the gala/wingpanel branches :)

PR #711, opened 2024-03-07
Port to GTK 4

### LINKED ISSUE #588: Port Greeter to GTK4

danirabbit:
Fixes #588 

- [ ] invalid cast from 'GtkText' to 'GtkEntry'
- [ ] logged in check is missing
- [ ] Doesn't show window in greeter session (shows when run in normal session)


lenemter:
> Doesn't show window in greeter session (shows when run in normal session)

This is an issue with mutter's WaylandClient. We have to wait for OS 9 and then start launching Greeter and other shell clients via systemd (or in case of Greeter maybe some other way)

PR #713, opened 2024-03-07
SessionButton: build menu from model

danirabbit:
No access to ModelButton in GTK 4


jeremypw:
I tried running `lightdm --test-mode --debug` in `Xephyr` as  suggested by the README but `main` did not display properly. Is there a better way to test?

danirabbit:
You can also run it directly, but just be aware that it'll replace your compositor and probably mess up your wallpaper

lenemter (approved):
The popover has margins now but other than that it works fine

PR #716, opened 2024-03-08
MainWindow: Use EventControllerKey

danirabbit:
* Replace key press event with key controller
* Replace get_style_context with state flags while we're here. The only non-deprecated use of stylecontext is to add a provider to the display


lenemter (changes requested):
This branch blocks Enter key, pressing it no longer logins you in

zeebok (approved):
This seems to work as intended with I tested in an OS8 VM!

PR #726, opened 2024-05-14
Make the compositor be the first program to launch

tintou:
Allows to make sure that the greeter will be launched when the compositor is ready, also allow us to use the right API on wayland.

### DIFF
--- a/compositor/WindowManager.vala
+++ b/compositor/WindowManager.vala
@@ -51,8 +51,6 @@ namespace GreeterCompositor {
 
         public GreeterCompositor.SystemBackground system_background { get; private set; }
 
-        private Clutter.Actor fade_in_screen;
-
         private Meta.PluginInfo info;
 
         // Used to toggle screenreader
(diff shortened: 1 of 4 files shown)


danirabbit:
Yay thank you! I keep meaning to look into this so that's really appreciated 🩷

tintou:
Anyone can feel free to take over this MR as I only have limited time currently 👍

tintou:
It is actually working when removing the switch to Gtk.Application so yay!

compositor/WindowManager.vala
```diff
@@ -85,12 +83,6 @@ namespace GreeterCompositor {
 
             disable_tiling_shortcuts ();
 
-            fade_in_screen.save_easing_state ();
-            fade_in_screen.set_easing_duration (1000);
-            fade_in_screen.set_easing_mode (Clutter.AnimationMode.EASE);
-            fade_in_screen.opacity = 0;
-            fade_in_screen.restore_easing_state ();
```
danirabbit, line 92: Was there a reason for removing the fade in? This branch seems to still work with it

danirabbit (commented):
Can confirm it does work!

For some reason my install has started not starting the panel in either the greeter or the main session :thinking: so I don't know what's the cause. But also I'm not getting the wallpaper on the greeter anymore either. I can't confirm if it's related to this branch unfortunately.

But, this does seem to solve a major issue where sometimes the screen would just be black or only partially faded in. So functionally this is a big fix. I'd be in favor of merging and figuring out those regressions separately if anyone else is experiencing them.

Just one inline comment :)

danirabbit (approved):
Reverted changes to fadein. If that's important for something please propose it again separately :)

PR #729, opened 2024-05-16
Make datetime widget larger

danirabbit:
Make the datetime widget larger, bolder, with softer shadows, and more towards the center of the display

## AFTER
Screenshot from 2024-05-16 13 35 25

## BEFORE
Screenshot from 2024-05-16 13 38 02

### DIFF
--- a/data/DateTime.css
+++ b/data/DateTime.css
@@ -1,19 +1,24 @@
 .date,
 .time {
     color: #fff;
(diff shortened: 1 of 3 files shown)


zeebok (approved):
Code looks good. I personally think this is an improvement so it gets my approval. UX should chime in obviously.

PR #733, opened 2024-05-25
MainWindow: Better detect when running in live session

### LINKED ISSUE #732: Login card is displayed in the installer session
### What Happened?

Login card (username and password input) is displayed under the installer window if I boot from the elementary OS 8 Early Access ISO.

### Steps to Reproduce

1. Boot from ISO (I used `elementaryos-8.0-daily.20240525.iso`)
2. Wait until installer window is shown
3. Move the installer window with Super + drag
4. See the login card of Live session user appears

image1

iamge2

### Expected Behavior

Login card should not be displayed.

### OS Version

8.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

danirabbit:
Fixes an issue where if you install Installer in a regular session that you'll break your lock screen

Maybe fixes #732 ?

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -36,7 +36,27 @@ public class Greeter.MainWindow : Gtk.ApplicationWindow {
     private int current_user_card_index = 0;
     private unowned Greeter.BaseCard? current_card = null;
 
-    private bool installer_mode = false;
+    private bool _is_live_session? = null;
+    private bool is_live_session {
+        get {
+            if (_is_live_session != null) {
+                return _is_live_session;
+            }
+
+            var proc_cmdline = File.new_for_path ("/proc/cmdline");
+            try {
+                var dis = new DataInputStream (proc_cmdline.read ());
+                var line = dis.read_line ();
+                if ("boot=casper" in line || "boot=live" in line || "rd.live.image" in line) {
+                    return true;
+                }
+            } catch (Error e) {
+                critical ("Couldn't detect if running in Live Session: %s", e.message);
+            }
+
+            return false;
+        }
+    }
 
     private Gtk.EventControllerKey key_controller;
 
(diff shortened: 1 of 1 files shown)


src/MainWindow.vala
```diff
@@ -36,7 +36,27 @@ public class Greeter.MainWindow : Gtk.ApplicationWindow {
     private int current_user_card_index = 0;
     private unowned Greeter.BaseCard? current_card = null;
 
-    private bool installer_mode = false;
+    private bool _is_live_session? = null;
+    private bool is_live_session {
+        get {
+            if (_is_live_session != null) {
+                return _is_live_session;
+            }
```
ryonakano, line 44: This looks like a dead code since `_is_live_session` is never substituted.

src/MainWindow.vala
```diff
@@ -36,7 +36,27 @@ public class Greeter.MainWindow : Gtk.ApplicationWindow {
     private int current_user_card_index = 0;
     private unowned Greeter.BaseCard? current_card = null;
 
-    private bool installer_mode = false;
+    private bool _is_live_session? = null;
```
ryonakano, line 39:
```suggestion
    private bool? _is_live_session = null;
```

CI complains this syntax error.

ryonakano (changes requested):
Thank you for your quick work! I'm not sure how to confirm this fixes the issue, but reviewed the code.

src/MainWindow.vala
danirabbit: I should maybe make sure I don’t code before having coffee 😅

danirabbit:
@ryonakano sorry for the mistakes! I just woke up haha thanks for the review :)

ryonakano (approved):
I'm not sure if this fixes #732 but I confirmed the issue installing the Installer breaks login screen no longer happens. Maybe we can check for #732 later when the ISO is built

PR #735, opened 2024-05-30
MainWindow: fallback to checking for installer if proc/cmdline fails

### LINKED ISSUE #734: Installer window is not shown
### What Happened?

Installer window is not shown in the live session and just the wingpanel is shown:

screen

### Steps to Reproduce

1. Boot from Early Access ISO (`elementaryos-8.0-daily.20240527.iso` is not reproducible but `elementaryos-8.0-daily.20240529.iso` is reproducible)
2. Wait until greeter launches
3. See Installer window is not shown

### Expected Behavior

Installer window is shown.

### OS Version

8.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

danirabbit:
In #733 we switched methods for checking for proc/cmdline. Add a fallback to the old method of looking for the installer in case that doesn't work for some reason

Also go ahead and select that default session if it exists

Might fix #734 :woman_shrugging:

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -53,7 +53,13 @@ public class Greeter.MainWindow : Gtk.ApplicationWindow {
                     _is_live_session = true;
                 }
             } catch (Error e) {
-                critical ("Couldn't detect if running in Live Session: %s", e.message);
+                critical ("Couldn't read proc/cmdline: %s", e.message);
+
+                // Check if the installer is installed
+                var installer_desktop = new DesktopAppInfo ("io.elementary.installer.desktop");
+                if (installer_desktop != null) {
+                    _is_live_session = true;
+                }
             }
 
             return _is_live_session;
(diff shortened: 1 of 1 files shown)


danirabbit:
I don't see anything in journalctl that would imply that this is failing, but doesn't hurt to have a fallback. It's more likely the issue is that select session bit

ryonakano:
This didn't worked for me unfortunatelly:

screen

FYI I'm testing with the following steps:

1. Install this branch and the latest Installer main branch
2. Fake the /proc/cmdline with the following command: `cat /proc/cmdline | tr -d '\n' > /tmp/cmdline; echo " boot=casper" >> /tmp/cmdline && sudo mount -B /tmp/cmdline /proc/cmdline`
3. Logout
4. Don't forget `sudo umount /proc/cmdline` after testing

ryonakano:
Closing since #733 is reverted in #736.

PR #747, opened 2024-08-08
Update 40-io.elementary.greeter.conf

danirabbit:
Set wayland as default session

### DIFF
--- a/data/40-io.elementary.greeter.conf
+++ b/data/40-io.elementary.greeter.conf
@@ -4,4 +4,4 @@ logind-check-graphical=true
 
 [Seat:*]
 greeter-session=io.elementary.greeter
-user-session=pantheon
+user-session=pantheon-wayland


danirabbit:
Should we go ahead and go Wayland by default for OS 8 or should we stick with X11 for another cycle?

zeebok:
Is there any major support issues with Wayland? I know for a long while nvidia simply didn't work with it but is there anything else to be concerned about?

danirabbit:
So far that I know of:

1. Some apps don't work well in the Wayland session. GIMP has issues with its menus. Boxes crashes the session (last I checked). Any app that isn't used portals for things like color picker isn't going to work

2. No onscreen keyboard. Onboard doesn't work in Wayland

3. Some Orca features don't work. I'm not sure which ones, this is just what I've heard

So maybe better to stick to X11 by default this cycle and continue to iron things out?

danirabbit:
See also https://github.com/orgs/elementary/projects/118

UncleTallest:
> Is there any major support issues with Wayland? I know for a long while nvidia simply didn't work with it but is there anything else to be concerned about?

98% of the Nvidia issues are fixed in driver ~v558 when esync was added to the driver. So before setting Wayland as default we should also set the default Nvidia driver to that version or higher.

zeebok:
I think that is pretty reasonable. Given distaste towards Windows 11 _seems_ to be growing, it would be good if we have a solid baseline of support for anything that is default in the system.

teamcons:
S a way for the greeter to fall back to starting in x11 mode if there are display issues for some nvidia peeps would be nice
Such as a keyboard shortcut
They then could change the session. But if they cant display the greeter at all, systems as good as boinked

danirabbit:
@teamcons the greeter itself still runs in X11. This is for changing the default user session, not the greeter session

PR #749, opened 2024-08-16
Add ShellClients support

### LINKED ISSUE #748: Add dock/panel positioning to greeter compositor
### Problem

The panel now depends on positioning being done in the compositor

### Proposal

Copy the panel positioning protocol stuff here from gala

### Prior Art (Optional)

_No response_

danirabbit:
Fixes #748 

This doesn't fix that the panel takes a hella long time to start in the Greeter for some reason, but it does fix the positioning!


leolost2605:
Not sure do we want the full thing here or only for wingpanel? If only wingpanel we can strip this down a whole lot.
(E.g. remove Barrier.vala, CenteredWindow.vala, HideTracker.vala, NotificationsClient.vala (This can and probably should be removed anyways) and a whole lot of code). I can do the strip if you want but I haven't worked with greeter compositor yet so I'm not sure about the integration. Although it seems to be just a frankensteined version of gala I guess :)

Also which config file is used here to start the wayland clients when running under wayland? (Does greeter even support wayland yet :grimacing: ?)

danirabbit:
@leolost2605 Yeah I was wondering if it made sense to strip it down or if it was better to keep it so we can keep it in sync with Gala updates? I think we want notifications for the volume and brightness confirmations right?

I just hardcoded the panel here in the area of which things to launch. But no I don't think the greeter session supports wayland? I think we need to keep it X11 for a11y

danirabbit:
@leolost2605 I removed Barrier and HideTracker since those are unused by the panel, but I left in Center and Notifications because I'd like to use center for installer and initial setup and it's used for the session dialog and I think we do want notifications for indicator confirmations at least

leolost2605:
> I think we want notifications for the volume and brightness confirmations right?

Ah right I forgot about that :sweat_smile: 

With the removal of HideTracker I guess we probably should also remove the corresponding protocol methods?
Apart from that I guess this should be fine but maybe someone who knows the whole greeter structure and stuff better than me can confirm :)

danirabbit:
I guess that would be maybe @tintou or @lenemter ?

lenemter:
@danirabbit I don't think we should modify the protocol file (protocol/pantheon-desktop-shell-v1.xml) here in any way. Instead we can just ignore the options we don't require or use yet.

danirabbit:
@lenemter done!

lenemter (approved):

PR #750, opened 2024-08-24
Use new gala dbus name

lenemter:
Also remove old greeter dbus code


danirabbit:
Drafting until https://github.com/elementary/gala/pull/2046 is merged

PR #752, opened 2024-08-27
Use a wayland session to start the greeter

tintou:
Opened as a draft until we are fully compatible.


tintou:
This can probably be merged at the same time #747 is considered ready

danirabbit:
My understanding is that Orca is missing some features under Wayland still and there might be other a11y tools that aren't fully working. So I think that until Newton/AccessKit lands we might be stuck with X11 on the greeter so that people who need those accommodations are able to choose a session

meisenzahl:
onboard does not work under Wayland either: https://github.com/elementary/session-settings/issues/82

danirabbit (approved):

PR #756, opened 2024-09-13
Fix Wingpanel DBus interface implementation

lenemter:
Same changes as https://github.com/elementary/wingpanel/pull/557


danirabbit:
In this branch I get wrong panel colors on startup but if you switch wallpapers and then switch back it'll fix the colors

PR #761, opened 2024-09-30
Main: don't start settings daemon in installer session

danirabbit:
In an installer session, settings daemon is started by the session manager. Should also fix gsettings overrides not working (for example for network notifications)


ryonakano:
I tested this branch by creating [a hock](https://github.com/ryonakano/os/commit/d01f4c6f1f087e72d54a80c21a136a2f35e352ef) to install the greeter .deb while building the iso, but unfortunately this does not fix https://github.com/elementary/installer/issues/775 :cry:

VirtualBox_test_01_10_2024_21_10_38

VirtualBox_test_01_10_2024_21_04_45

danirabbit:
@ryonakano hm okay. Still probably shouldn't be starting GSD twice here. I also moved in a bit about settings the session name which affects gsettings overrides

ryonakano:
I installed this branch and Installer and then see the background of Greeter is now blue instead of blurred wallpaper. Is this an expected behavior?

VirtualBox_elementary OS 8_08_10_2024_07_39_39

danirabbit:
Ah no it's not :sweat_smile:

danirabbit:
@ryonakano did that fix translations though?? Or is this because you logged in to demo session to install it?

ryonakano:
@danirabbit Unfortunately the translation issue still presents. I booted from the ISO built using this branch and it looks like this:

VirtualBox_test_08_10_2024_21_11_56

danirabbit:
@ryonakano Wallpaper issue should be fixed by https://github.com/elementary/installer/pull/840

ryonakano:
Built a daily ISO with this branch and confirmed the wallpaper issue has been resolved.

@danirabbit This PR looks good to me, but what is the difference with #768?

danirabbit:
@ryonakano that other PR was just me testing. It can be closed

ryonakano (approved):
I see, let's merge this PR then. Sorry for my late review 🙏

PR #779, opened 2025-01-28
Run our session manager as a separate process

lenemter:


danirabbit (approved):
Fixes the linked issues. Nice job!

PR #781, opened 2025-01-28
Compositor: remove more dead code

lenemter:


leonardo-lemos (approved):
LGTM. Works without any regression.

Issue #688, opened 2023-10-12
Pantheon greeter (login screen and lock screen) on elementary OS 7.1 does not follow the dark visual style


eric0001:
### What Happened?

![login_2](https://github.com/elementary/greeter/assets/39679934/c2c39328-3dae-4238-9801-ccccc7374104)

![login_1](https://github.com/elementary/greeter/assets/39679934/0e885722-b3a3-4463-acaa-0da5741b2e9f)

### Steps to Reproduce

Go to system settings in elementary OS 7.1 under Desktop and Appearance.
Set the preferred visual style to Dark.
Log out.
On the login screen, the login card does not follow the dark style i.e. the login card has a light background.

### Expected Behavior

One would expect to have a login card with a dark background just like any other dialog or window in dark style.

### OS Version

7.x (Horus)

### Software Version

Latest release (I have run all updates)

PR #563, opened 2021-09-06
Greet Now Support Dark_Mode 

BAProductions:
Greet Now Support Dark_Mode Code is based on merge request [#499](https://github.com/elementary/greeter/pull/499)

PR #499, opened 2021-01-03
[WIP] Use the users dark style/accent color preference

fs3o5:
This PR sets the selected users theme preferences.

Known problems:
- [ ] Wingpanel does not respect the theme choices.
- [ ] https://github.com/elementary/switchboard-plug-pantheon-shell/issues/253 must be implemented.

PR #786, opened 2025-05-09
Add dark style support

lenemter:
Fixes https://github.com/elementary/greeter/issues/688
Closes #563
Closes #499

This branch implements mini settings portal inside of the greeter itself and exposes it via dbus which allows granite to correctly get dark style preference


leonardo-lemos (approved):
Works as expected, nice job!

lenemter:
It should respect dark style schedule now. It would be awesome if more people tested whether this approach works on their machine :)

danirabbit (approved):
This is a very interesting approach! Works perfectly for me

danirabbit:
Doesn't seem like anyone has objected to this so let's merge it for testing in Daily

PR #796, opened 2025-06-08
Push Greeter to the background using pantheon shell protocol

lenemter:
Wayland and Gtk4 prep


leonardo-lemos (approved):
Works without regression AFAIK

PR #802, opened 2025-06-09
Connect to lightdm daemon early

lenemter:
Allows us to get rid of these ugly notifies

It doesn't make sense to connect to daemon asynchronously because the window is invisible until the users are loaded, which is when we connect to the daemon

While we're here use box for the bottom buttons and don't store them in the main window variables

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -190,7 +190,6 @@ public class Greeter.MainWindow : Gtk.ApplicationWindow {
 
             if (children.nth_data (index) is Greeter.UserCard) {
                 current_user_card_index = (int) index;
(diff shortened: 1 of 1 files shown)


src/MainWindow.vala
```diff
@@ -199,7 +190,7 @@ public class Greeter.MainWindow : Gtk.ApplicationWindow {
 
             if (children.nth_data (index) is Greeter.UserCard) {
                 current_user_card_index = (int) index;
-                switch_to_card ((Greeter.UserCard) children.nth_data (index));
+                //  switch_to_card ((Greeter.UserCard) children.nth_data (index));
```
leonardo-lemos, line 193: Commented out code.

leonardo-lemos (approved):

Issue #778, opened 2024-12-25
lost HAS_MUTTER46 ?


tumashu:
https://github.com/elementary/greeter/blob/084ab7c1e3383bea177aa3e67bfeecfacad9c107/compositor/ShellClients/ShellClientsManager.vala#L63

jgarte:
Hi, what is the issue that you ran into here? 

It's a bit unclear what this issue is about.

tumashu:
mutter < 46 have no make_dock function, so it will build fail when mutter < 46.

tumashu:
like this:
```
+#if HAS_MUTTER46
                 client.wayland_client.make_dock (window);
+#endif
```

PR #810, opened 2025-06-10
Compositor: sync shell clients code with gala

lenemter:
Fixes https://github.com/elementary/greeter/issues/778


tintou (approved):
Untested but the code looks conform!

PR #824, opened 2025-09-14
Do not try to fallback to x11 session when it is unavailable

bobby285271:
- Motivation: elementary/session-settings#91


danirabbit (approved):
Thank you!

PR #828, opened 2025-10-03
Simplify disabling tiling shortcuts

lenemter:


danirabbit (approved):
Nice simplification. All works as expected afaict

Issue #831, opened 2025-10-31
Switching cards with the keyboard never reveals the password field


o-alquimista:
### What Happened?

When I try to switch focus to another card and type the password, the view moves horizontally but it doesn't show the card's password field. Clicking with the mouse works, though.

https://github.com/user-attachments/assets/25bebd8d-d20f-45ea-ace1-6bc024f54152

Things I've tried:
- Move to card and press Enter.
- Move to card and start typing.

### Steps to Reproduce

1. Create two or more users.
2. Login as one user, then lock the screen to get to the greeter.
3. Attempt to login as another user using the left/right arrow keys on the keyboard.

### Expected Behavior

When you switch cards with the keyboard, the password field of the card in focus should appear and be ready to receive the password.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell

```

PR #832, opened 2025-11-07
Fix card switching

lenemter:
Fixes https://github.com/elementary/greeter/issues/831

Use `carousel.scroll_to (...)`, instead of `switch_to_card`. And rename `switch_to_card` to `handle_page_changed`
Another change is that I removed `UserCard.focus_requested` signal and moved handling card clicks to the MainWindow


danirabbit (approved):
This definitely seems more reliable and it works as far as I can tell and I can't find a way to break it :)

PR #835, opened 2025-11-08
Remove accountservice dependency

lenemter:
LightDM already provides everything we need


danirabbit:
Oh that's fantastic. Do we need to bump our lightdm dep version?

PR #839, opened 2025-11-10
Remember session type per-user

### LINKED ISSUE #830: Remember session type (Secure or Classic) per-user
### Problem

A little context. I use the Secure Session for everything, except for competitive gaming. That is because the Wayland compositor (Mutter) has not yet implemented the [tearing control protocol](https://wayland.app/protocols/tearing-control-v1), which was made to let games output to the screen as fast as they can without having to synchronize with the monitor refresh rate — which is perceived as screen tearing. This is especially important when playing a competitive first-person shooter such as CS:2. There is a small but noticeable delay in-game that makes your mouse movements feel slower — like it's heavy, or dragging behind.

My workaround to this problem is to **create a separate user** just for running some Steam games on a Classic session. My main user should continue to log in to a Secure session. Unfortunately, **the elementary greeter does not remember the session type as a user preference**, and it proceeds to load whatever the last user chose.

Screenshot of the Greeter showing cards for my user and a Steam user

### Proposal

**Store the last session type used by each user as a preference** somewhere. That way, a user would just log in to whatever session they prefer, without pissing off other users. Since the session type selector is contained within each user card in the greeter, it is natural to assume it is different for each user.

### Prior Art (Optional)

_No response_

lenemter:
Fixes https://github.com/elementary/greeter/issues/830

Turns out lightdm already remembers this so this is pretty easy to implement but required a lot of code to replace a single application action with an action for each of the cards.

I'm not experienced with actions so if there's an easier way to implement this please let me know


zeebok:
> Are we going to even offer Classic session in OS9?

As far as I know, we are dropping it for OS9 *but* given there is no real sign of a release date yet, it can't hurt to get this in for 8.1

PR #870, opened 2026-03-28
Control: update for Mutter 18

danirabbit:
Goes with #867


danirabbit:
@ryonakano thank you for catching that! I suppose that means we'll need to disable noble builds of Greeter when this is merged or create a noble packaging branch

PR #873, opened 2026-04-23
Add systemd notify

danirabbit:
Goes with https://github.com/elementary/installer/pull/929

Like https://github.com/elementary/gala/pull/1256, but the systemd units are created in the installer repo for some reason? Should I put those here instead?


lenemter (approved):
LGTM

danirabbit:
Closing since we're using Gala as the compositor

PR #875, opened 2026-04-24
PantheonShell: match function order in Gala

danirabbit:


danirabbit:
Closing since we're using Gala as the compositor

PR #878, opened 2026-04-25
Use gala instead of the greeter compositor

### LINKED ISSUE #871: Wayland compositor launcher
### Problem

We need a wayland session for installer but there's no wayland launcher for the greeter compositor

### Proposal

Create a wayland launcher

### Prior Art (Optional)

Gala

leolost2605:
Fixes #871

### DIFF
--- a/data/io.elementary.greeter.desktop.in
+++ b/data/io.elementary.greeter.desktop.in
@@ -2,7 +2,7 @@
 Name=Pantheon Greeter
 Comment=Pantheon Greeter
 Icon=io.elementary.greeter.settings
-Exec=@PROJECT_NAME@-compositor
+Exec=env GALA_SESSION_TYPE=greeter gala --wayland
 Type=Application
 X-LightDM-Session-Type=wayland
 NoDisplay=true
--- a/data/io.elementary.greeter.desktop.in
+++ b/data/io.elementary.greeter.desktop.in
@@ -2,7 +2,7 @@
 Name=Pantheon Greeter
 Comment=Pantheon Greeter
 Icon=io.elementary.greeter.settings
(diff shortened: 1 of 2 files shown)


lenemter:
@leolost2605 Do you mind merging this?

leolost2605:
I would be fine with merging though there are some things I'm unsure about:
- I haven't tested the installer path at all
- Is there something that we would lose support for when the greeter runs on wayland and not x?
- There are still some minor issues (e.g. wingpanel black bar, and now with making the greeter a desktop window I noticed you have to click it once before it is focused) but they should be easy fixes

Those questions are ofc only relevant if we want to make another greeter release for OS8, do we?

lenemter:
@leolost2605 The installer path is untouched in this branch, and releasing Greeter to OS 8 is problematic, since Greeter already runs on Wayland since #752

AFAIK installer session runs greeter-compositor and launches Installer via autostart file, without the need for Greeter, so I guess it's fine

leolost2605:
> @leolost2605 The installer path is untouched in this branch, and releasing Greeter to OS 8 is problematic, since Greeter already runs on Wayland since https://github.com/elementary/greeter/pull/752

Ahh well then we can merge from my side :)

> AFAIK installer session runs greeter-compositor and launches Installer via autostart file, without the need for Greeter, so I guess it's fine

Any idea where the files are located for that? I had a quick look through the installer repo but didn't find anything :(
I'd like to test it at some point when I have time

lenemter:
It looks like the installer session files are stored here: https://github.com/elementary/installer/tree/main/session

lenemter (approved):

Issue #868, opened 2026-03-12
greeter requires both a password and finger print to log in


SammyPants37:
### What Happened?

when attempting to log in on system boot, I am required to first input my password and then touch my fingerprint sensor to log in successfully. If I do not touch my fingerprint reader within the timeout period, I am given a generic login error and am told to try again. If I only touch my fingerprint reader without entering the password, nothing happens.
This happens on my fresh arch install that I've made sure not to install too many things on.

### Steps to Reproduce

add a finger to fprintd
add fprintd as a sufficient method of auth to /etc/pam.d/system-auth
try to log in on first boot (lock screen works fine but I'm pretty sure that is a different service)

### Expected Behavior

greeter should let you to either use your password or use your fingerprint not both (and especially not require both and not say anything about needing the fingerprint)

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland, This is the default)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
relevant logs from journalctl with the word greeter in them

Mar 12 10:42:01 Sammypants-Laptop greetd[756]: config: Config { file: ConfigFile { terminal: ConfigTerminal { vt: Specific(1), switch: true }, general: ConfigGeneral { source_profile: true, runfile: "/run/greetd.run", service: "greetd" }, default_session: ConfigSession { command: "dms-greeter --command niri -C /etc/greetd/niri/config.kdl", user: "greeter", service: "greetd-greeter" }, initial_session: None }, internal: ConfigInternal { session_worker: 0 } }
Mar 12 10:42:01 Sammypants-Laptop greetd[757]: pam_unix(greetd:session): session opened for user greeter(uid=964) by greeter(uid=0)
Mar 12 10:42:01 Sammypants-Laptop systemd-logind[676]: New session '1' of user 'greeter' with class 'greeter' and type 'tty'.
Mar 12 10:42:01 Sammypants-Laptop (systemd)[768]: pam_warn(systemd-user:setcred): function=[pam_sm_setcred] flags=0x8002 service=[systemd-user] terminal=[] user=[greeter] ruser=[<unknown>] rhost=[<unknown>]
Mar 12 10:42:01 Sammypants-Laptop (systemd)[768]: pam_unix(systemd-user:session): session opened for user greeter(uid=964) by greeter(uid=0)
[...]

Issue #738, opened 2024-06-17
Login window no longer responds after password entry in combination with alternative login methods


lug-gh:
### What Happened?

If you activate other login methods in addition to the password (e.g. biometric), it happens from time to time that the Greeter freezes when you enter your password. The "password characters" then turn a different shade of gray and the password line goes into read-only mode. It is also not possible to press Login or [Enter]. The buttons at the top of the screen work, however, and if you switch to another monitor with the mouse, the Greeter also switches with it.

If you switch to a different tty, you can log in and view log files, for example, but if you switch back to the desktop environment, you only have a mouse pointer and the greeter is no longer visible.

### Steps to Reproduce

Enroll some alternative Login Method, e.g. fingerprint or face-unlock. Then simply work with the device for a while and lock the screen from time to time. I can't really reproduce it, as it only occurs sometimes.

### Expected Behavior

Login should work fine

### OS Version

7.x (Horus)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
I can't find any Error Logs regarding this issue, but if someone can guide me for what to look, i can upload the logs.
```

### Hardware Info

Dell Latitude 5431

JoeMcNuggets:
Same on the 8.0

Issue #494, opened 2020-12-17
improve user feedback when unlock with fingerprint


andirsun:
## Problem

Currently the greeter is unlocking correctly with my fingerprint sensor but no exists any confirmation check when the laptop unlock

## Proposal

Show a confirmation icon like ```emblem-default``` when the fingeprint sucess the check.

ibnesayeed:
My experience goes like this:

* Greeter does not show any signs of fingerprint auth initially
* First attempt of password auth (with correct password) fails
* Fingerprint icon appears after that
* Fingerprint auth works

PR #884, opened 2026-07-04
Fast loop fprint auth to simulate simultaneous nonblocking logins with passwd auth

### LINKED ISSUE #502: Allow fingerprint login together with password login
## Problem

Currently if fingerprint login is enabled via installing the `fprintd` package and editing the `/etc/pam.d/lightdm` file to add `auth sufficient pam_fprintd.so` line then the greeter does not allow you to login by entering your password until the fingerprint module times out.

## Proposal

Allow users to login via fingerprint swipe or via entering their password whichever one is correctly attempted first.

## Prior Art

Stock ubuntu focal 20.04 gdm login screen lets you login using either fingerprint or entering password - one does not block the other.

vjr:
Promising workaround to enable (simulate) fprint login to work "simultaneously" with passwd login without one "blocking" the other.

Requires the following line added at the top of `/etc/pam.d/lightdm` but this branch works (passwd-only login) without it. This file is owned by the `lightdm` package not elementary OS so unsure how or where to add this in code/config/packaging or just leave it to users to add it themselves:

`auth sufficient pam_fprintd.so max-tries=1 timeout=1`

Note in Ubuntu 24.04 LTS (elementary OS v8) the fprintd timeout is minimum 10 seconds which seems to be fixed/works in Ubuntu 26.04 LTS (for elementary OS v9) which means in os8 when a user enters password has to wait up to 10 seconds before login happens but in os9 likely wait for a maximum of 1 second.

Needs some testing on Ubuntu 26.04 LTS to validate this short timeout does not cause fprint to fail due to insufficient duration for the device to verify a fingerprint!

edit: Tested builtin fprint reader on laptop and a USB reader and fprint verification works with this lowered 1-second timeout.

Fixes https://github.com/elementary/greeter/issues/502

Addresses https://github.com/elementary/greeter/issues/868
Addresses https://github.com/elementary/greeter/issues/738
Addresses https://github.com/elementary/greeter/issues/494


vjr:
Fixes #502

Addresses #868
Addresses #738
Addresses #494

PR #885, opened 2026-07-06
Greeter: avoid early authentication

### LINKED ISSUE #413: Buggy no-password user logout
I have 3 user: The administrator one with password, and the other 2 are standard without password (I eliminated the password by `sudo passwd -d username`).
If I'm logged in a no-password user and I logout, the screen goes black (It goes back to the login screen) and after that it logs in the same account. That's because if you select a user (Whithout password) in the login screen, it logs in inmediatly.
When you log out, that account is represented as the selected one in the login screen, so if you log out with a no-password user, it selects that one, and it logins in that one inmediatly.

lenemter:
Fixes https://github.com/elementary/greeter/issues/413

Current workflow in main is "carousel page switched" -> "authenticate by username" -> "wait for password" -> "respond with password". But when a user has no password set, `LightDM.Greeter.authenticate (username)` logs the user in without the need to provide the credential via `LightDM.Greeter.respond (credential)`.

This PR changes workflow to "Wait for a card to call `authenticate` with username and credential" -> "authenticate" -> "respond"

Also I put this logic into a separate class since `MainWindow.vala` is very long and it's very hard to read code there.

### DIFF
--- a/src/AuthenticationManager.vala
+++ b/src/AuthenticationManager.vala
@@ -1,0 +1,63 @@
+/*
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ * SPDX-FileCopyrightText: 2026 elementary, Inc. (https://elementary.io)
+ *
+ * Authors: Leo "lenemter" <lenemter@gmail.com>
+ */
+
+private class Greeter.AuthenticationManager : Object {
+    public LightDM.Greeter lightdm_greeter { private get; construct; }
+
+    private bool waiting_for_authentication_to_complete = false;
+    private string queued_credential = "";
+
+    public AuthenticationManager (LightDM.Greeter lightdm_greeter) {
+        Object (lightdm_greeter: lightdm_greeter);
+    }
+
+    construct {
+        lightdm_greeter.show_prompt.connect (on_show_prompt);
+    }
+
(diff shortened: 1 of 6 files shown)


lenemter:
Converting to draft to think about how this will affect fingerprint auth

vjr:
> Converting to draft to think about how this will affect fingerprint auth

Hi! I think you can go ahead with this PR and not worry about fprint auth which is already currently quite broken. As long as a user can still add the fprintd PAM config and "somewhat" login with their fprint in this PR we should be good IMO. My PR #884 to address/workaround fprint auth/login to get it in an acceptable condition I can rework the code after your PRs get merged, it's a small diff anyway, assuming reviewers don't raise concerns with the approach. Thanks!

lenemter:
@vjr That's the issue. I'm afraid fingerprint auth might not work at all in this branch because greeter doesn't start authentication until the password is entered

BAProductions:
> @vjr That's the issue. I'm afraid fingerprint auth might not work at all in this branch because greeter doesn't start authentication until the password is entered

Well that's how it should be fingerprint should not work on the greeter after reboot or boot i add message like MacOS does to say can't use until user login for better security

Issue #24, opened 2017-04-12
Allow for newly created users to set a password.


elementaryBot:
When a user is created without a password, they need to be able to set the new password prior to logging in. This needs to be supported in the greeter.

Launchpad Details: [#LP1405259](https://bugs.launchpad.net/bugs/1405259) Sam Hewitt - 2014-12-23 19:01:10 +0000

elementaryBot:
To be clear, this isn't just creating a user without a password. There is actually a feature where you create a new user with the intention of having them create their own password.

Launchpad Details: [#LPC](https://bugs.launchpad.net/pantheon-greeter/+bug/1405259/comments/2) Daniel Fore - 2014-12-23 19:07:24 +0000

PR #887, opened 2026-07-08
Use AccountsService for user info

lenemter:
Move away from LightDM for user info, as it is incomplete in many places

Will allow us to eventually fix issues https://github.com/elementary/greeter/issues/24 and https://github.com/elementary/greeter/issues/413

### DIFF
--- a/src/Cards/UserCard.vala
+++ b/src/Cards/UserCard.vala
@@ -1,12 +1,16 @@
 /*
- * Copyright 2018-2025 elementary, Inc. (https://elementary.io)
+ * Copyright 2018-2026 elementary, Inc. (https://elementary.io)
  * SPDX-License-Identifier: GPL-3.0-or-later
  *
  * Authors: Corentin Noël <corentin@elementary.io>
  */
 
 public class Greeter.UserCard : Greeter.BaseCard {
-    public LightDM.User lightdm_user { get; construct; }
+    /**
+     * We use Act.User instead of LightDM.User because lightdm is unmaintained
+     * and lacks some fields from Act.User such as `password_mode`.
+     */
+    public Act.User user { get; construct; }
     public bool show_input { get; set; default = false; }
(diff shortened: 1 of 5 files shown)


src/Cards/UserCard.vala
```diff
@@ -5,8 +5,8 @@
  * Authors: Corentin Noël <corentin@elementary.io>
  */
 
-public class Greeter.UserCard : Greeter.BaseCard {
-    public LightDM.User lightdm_user { get; construct; }
+private class Greeter.UserCard : Greeter.BaseCard {
+    public Act.User user { get; construct; }
```
danirabbit, line 13: Maybe it would be good to add a comment about why we're using Act instead of lightdm so that someone in the future doesn't think "oh let's make this use lightdm so we can drop a dependency" 😅

lenemter: Done

src/Cards/UserCard.vala
```diff
@@ -15,29 +19,32 @@ public class Greeter.UserCard : Greeter.BaseCard {
...
-        username_label = new Gtk.Label (lightdm_user.display_name) {
+        var username_label = new Gtk.Label (user.real_name) {
             hexpand = true,
             margin_top = 24,
             margin_bottom = 12,
             margin_start = 24,
             margin_end = 24,
         };
         username_label.get_style_context ().add_class (Granite.STYLE_CLASS_H2_LABEL);
+        user.bind_property ("locked", username_label, "sensitive", SYNC_CREATE);
```
danirabbit: This is inverted :) 
```suggestion
        user.bind_property ("locked", username_label, "sensitive", SYNC_CREATE | INVERT_BOOLEAN);
```

danirabbit (changes requested):
This looks good just one little oops

danirabbit (approved):

PR #890, opened 2026-07-08
Refactor authentication

lenemter:
Split auth into 2 steps: authentication and providing credentials. And make UserCard explicitly request auth start instead of doing it automatically in MainWindow


zeebok (approved):
LGTM

PR #894, opened 2026-07-25
Gtk4 prep

lenemter:


compositor/meson.build
```diff
@@ -3,6 +3,7 @@
 glib_dep = [dependency('gio-2.0'), dependency('gio-unix-2.0'), dependency('gmodule-2.0')]
 gtk_dep = [dependency('gtk+-3.0'), dependency('gdk-x11-3.0')]
 gee_dep = dependency('gee-0.8')
+gnome_desktop_dep = dependency('gnome-desktop-3.0')
```
danirabbit, line 6: Why did this dep move here? Aren't we removing the compositor soon?

lenemter: Greeter binary doesn't depend on gnome-desktop-3.0, so I moved the dependency to compositor's meson file

danirabbit (approved):

PR #897, opened 2026-07-27
Update packaging to use gala and gtk4

lenemter:


danirabbit (approved):

debian/control
```diff
@@ -25,6 +26,7 @@ Homepage: https://github.com/elementary/greeter
 Package: io.elementary.greeter
 Architecture: any
 Depends: io.elementary.settings-daemon (>=8.4.0),
+         gala
```
danirabbit: Missing comma :)
```suggestion
         gala,
```

danirabbit (changes requested):

debian/control
lenemter: Oops

danirabbit (approved):

